PR #4, opened 2021-03-15
Watch dev script and start simulators

dagda1:
## Motivation

Add a watch script for development and start the simulator express servers.

## Approach

Used affection and started the simulators imperatively in the `SimulationContext`

### Alternate Designs

When we have the `Atom` we could start the simulators when they are added to the state.

### DIFF
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -30,6 +30,7 @@
     "assert-ts": "^0.3.2",
     "effection": "=2.0.0-preview.5",
     "express": "^4.17.1",
+    "get-port": "^5.1.1",
     "graphql": "^15.5.0",
     "nexus": "^1.0.0",
     "uuid": "^8.3.2"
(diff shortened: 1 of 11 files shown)


packages/server/package.json
```diff
@@ -28,12 +30,16 @@
...
+    "get-port": "^5.1.1",
     "graphql": "^15.5.0",
     "nexus": "^1.0.0",
     "uuid": "^8.3.2"
   },
   "devDependencies": {
+    "@effection/channel": "2.0.0-preview.4",
+    "@effection/events": "=2.0.0-preview.4",
     "@effection/mocha": "=2.0.0-preview.3",
+    "@effection/subscription": "=2.0.0-preview.4",
```
cowboyd, line 42: By having `effection` in the dependency graph, we should have access to all of these  since `effection` re-exports them. (with the exception of `@effection/mocha`

cowboyd (approved):

packages/server/package.json
dagda1: ok, definitely better to not manage these independently

PR #5, opened 2021-03-17
Extract http server logic into its own file

cowboyd:
## Motivation

One of the things we identified was that whenever you spawn resources into a scope like a server, it makes it much easier to reason about, and also to compose when the actual binding of the resource to a scope is a single, unary function. This is how streams work in that you can define a stream, and map it and modify it, but eventually you call `subscribe(scope)` on it to actually do the iteration. When necessary, you can define an operation on the stream like `forEach` that actually binds to the scope and performs the iteration so that you don't even need to ever call the `subscribe()` method.

We're [refactoring the `exec()` api](https://github.com/thefrontside/effection/pull/269) to follow this pattern. In effect, 

```js
exec(scope, command);
```

becomes

```js
exec(command).run(scope);
```

But more commonly, when you want to get the result of the command immediately, you would say:

```js
let result = yield exec(command).expect();
```
## Approach

This makes our http server creation follow this pattern, so before where we bound the scope and also returned an operation in a single step:

```js
yield spawnServer(scope, app);
```
We now explicitly bind servers to a scope in a single step, and then have a separate operation to wait until they are listening and get the port number

```js
let server = createServer(app).run(scope);

let { port } = yield server.listening();
```

### DIFF
--- a/packages/server/src/http.ts
+++ b/packages/server/src/http.ts
@@ -1,15 +1,13 @@
 import { Deferred, Operation, Task, once } from 'effection';
 import { Express } from 'express';
-import type { AddressInfo } from 'net';
 
+import type { AddressInfo } from 'net';
 export type { AddressInfo } from 'net';
+
+import type { Runnable } from './interfaces';
 
 export interface Server {
   listening(): Operation<AddressInfo>;
-}
-
-export interface Runnable<T> {
-  run(scope: Task): T;
 }
 
 export function createServer(app: Express): Runnable<Server> {
--- a/packages/server/src/http.ts
+++ b/packages/server/src/http.ts
@@ -1,0 +1,45 @@
+import { Deferred, Operation, Task, once } from 'effection';
+import { Express } from 'express';
+import type { AddressInfo } from 'net';
+
+export type { AddressInfo } from 'net';
+
+export interface Server {
+  listening(): Operation<AddressInfo>;
+}
+
+export interface Runnable<T> {
+  run(scope: Task): T;
+}
+
+export function createServer(app: Express): Runnable<Server> {
+  return {
+    run(scope: Task) {
+      let server = app.listen();
+
+      let bound = Deferred<void>();
+
+      scope.spawn(function*() {
+        try {
+          yield once(server, 'listening');
+          bound.resolve();
+          yield;
+        } finally {
+          server.close();
+        }
+      });
+
+      scope.spawn(function*() {
+        let error: Error = yield once(server, 'error');
+        throw error;
+      });
+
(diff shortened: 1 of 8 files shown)


packages/server/src/http.ts
```diff
@@ -0,0 +1,49 @@
...
+import { Express } from 'express';
+import type { AddressInfo } from 'net';
+
+export type { AddressInfo } from 'net';
+
+export interface Server {
+  listening(): Operation<AddressInfo>;
+}
+
+export interface Runnable<T> {
```
dagda1, line 11: should we put this in the `./interfaces` file it will be used in other places.

packages/server/src/http.ts
```diff
@@ -0,0 +1,49 @@
+import { Deferred, Operation, Task, once } from 'effection';
+import { Express } from 'express';
+import type { AddressInfo } from 'net';
+
+export type { AddressInfo } from 'net';
+
+export interface Server {
+  listening(): Operation<AddressInfo>;
```
dagda1, line 8: `listening` sounds like it should return a `boolean` but instead returns `AddressInfo`.

dagda1 (commented):
I definitely like that it does immediately start the server.   Nice work!

packages/server/src/http.ts
cowboyd: Makes sense 👍🏻

packages/server/src/http.ts
cowboyd: Hmmm, what do you think is a good name? I didn't want to call it `listen()` since that would imply it is at that point that we're binding the port, but the chances are good that the port is already bound when you call that method.

dagda1: > Hmmm, what do you think is a good name? I didn't want to call it `listen()` since that would imply it is at that point that we're binding the port, but the chances are good that the port is already bound when you call that method.

Might be just my opinion.  Happy to go with `listening` for now

packages/server/src/http.ts
```diff
@@ -0,0 +1,49 @@
...
+export function createServer(app: Express): Runnable<Server> {
+  return {
+    run(scope: Task) {
+      let server = app.listen();
+
+      let bound = Deferred<void>();
+
+      scope.spawn(function*() {
+        try {
+          if (server.listening) {
```
dagda1, line 24: if this is a method, should it not be `server.listening()`

dagda1: oh, this is this is the `http` server and not `our` server

dagda1 (approved):

PR #6, opened 2021-03-17
Add atom

dagda1:
## Motivation

We are going to need the atom at some stage but I was adding the auth0 simulator in another branch when I reached this point:

```
const jwksMock = createJWKSMock(`https://localhost:6467`);
```

This `jwksMock` object creates the mock access_tokens etc. that auth0 does, in order for this to work correctly, a certificate needs to be generated with the correct domain which would be the domain of the `auth0` https service or simulator.

## Approach

The approach in this PR is to use the atom to store the state of any service instances that have been spun up and are running on a specific port.

So this configuration:

```ts
      let { port } = yield spawnSimulationServer(world, atom, {
        simulators: {
          echo({ http }) {
            return http(app => app.get('/', echo));
          },
          secureEcho({ https }) {
            return https(app => app.get('/', echo));
          }
        }
      });
```

Would result in this state getting added:

```js
  serviceInstances: {
    'afec01e1-45c7-4bb8-914b-b97784051f90': {
      id: 'afec01e1-45c7-4bb8-914b-b97784051f90',
      port: 50648,
      name: 'echo',
      url: 'http://localhost:50648'
    },
    '03ef6472-08a6-41ab-b5d7-53e450691e16': {
      id: '03ef6472-08a6-41ab-b5d7-53e450691e16',
      port: 50649,
      name: 'secureEcho',
      url: 'https://localhost:50649'
    }
},
```
With this in place we should be able to use `once` 

```
simulator.slice.once((sim => sim.url)
```

### Alternate Designs

You `wait` for the http/https `listening` event to be raised

```
yield once(sim, 'listening');
```

### DIFF
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -25,6 +25,7 @@
   },
   "homepage": "https://github.com/thefrontside/simulacrum#readme",
   "dependencies": {
+    "@effection/atom": "^2.0.0-preview.3",
     "@effection/node": "=2.0.0-preview.5",
     "@simulacrum/client": "^0.0.0",
     "assert-ts": "^0.3.2",
--- a/packages/server/src/index.ts
+++ b/packages/server/src/index.ts
@@ -1,1 +1,1 @@
-export { spawnSimulationServer } from './server';
+export { spawnSimulationServer } from './server/server';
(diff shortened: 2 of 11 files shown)


cowboyd:
We ended up integrating the atom even more deeply with #9

PR #9, opened 2021-03-18
drive entire simulation as an effect mapped off simulation state

cowboyd:
Motivation
-----------
Since we're about the business of adding the atom to track simulation with https://github.com/thefrontside/simulacrum/pull/6 we have an opportunity to completely re-think how this will work and organize accordingly.

Approach
----------
The strategy is that the entire simulation is tracked in the atom from the moment it is created until the moment it is destroyed. Instead of doing everything in the graphql context, the context just inserts the initial simulation record into the atom and the `map()` effect which maps each record to an operation takes it away from there.

### DIFF
--- a/packages/server/src/effect.ts
+++ b/packages/server/src/effect.ts
@@ -1,0 +1,33 @@
+import { Slice } from "@effection/atom";
+import { Operation, Task } from "effection";
+
+export interface Effect<A> {
+  (slice: Slice<A>): Operation<void>;
+}
+
+export function map<A>(slice: Slice<Record<string, A>>, effect: Effect<A>): Operation<void> {
+  return function*(scope) {
+    let effects = new Map<string,Task>();
+
+    function synchronize(record: Record<string, A>) {
+      let keep = new Set<string>();
+
+      for (let key of Object.keys(record)) {
+        if (!effects.has(key)) {
+          effects.set(key, scope.spawn(effect(slice.slice(key))));
+        }
+        keep.add(key);
+      }
+
+      for (let [key, context] of effects.entries()) {
+        if (!keep.has(key)) {
+          context.halt();
+        }
+      }
+    }
+
+    synchronize(slice.get());
+
+    yield slice.forEach(synchronize);
+  };
+}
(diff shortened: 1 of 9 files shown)


packages/server/src/interfaces.ts
```diff
@@ -28,14 +28,31 @@ export interface Service {
...
-  id: string;
-  simulators: Record<string, Simulator>;
-  services: Service[];
-  addSimulator(name: string, simulator: Simulator): Simulation;
-  scope: Task;
+export interface ServerState {
+  simulations: Record<string, SimulationState>;
 }
 
+export type SimulationState =
```
dagda1, line 35: I love the smell of algebraic data types/discriminated untions in the morning :).  I was thinking along these lines also.

packages/server/src/effect.ts
```diff
@@ -0,0 +1,33 @@
+import { Slice } from "@effection/atom";
+import { Operation, Task } from "effection";
+
+export interface Effect<A> {
```
dagda1, line 4: this is seriously cool

packages/server/src/simulation.ts
```diff
@@ -0,0 +1,97 @@
...
+    services,
+    http(handler) {
+      let protocol: Protocols = 'http';
+      let app = handler(createHttpApp());
+      return createBehaviors(services.concat({ name, protocol, app }), name);
+    },
+    https(handler) {
+      let protocol: Protocols = 'https';
+      let app = handler(createHttpApp());
+      return createBehaviors(services.concat({ name, protocol, app }), name)
```
dagda1, line 76: does this cure the bug I mentioned previously where `services = []` every time `http` or `https` is called?

Might be worth having a test that adds more than 1 service and checks the length.

dagda1 (commented):
Although early, this codebase is already seriously nice :).

The only thing I would say about the PR is that it is light on tests

packages/server/src/simulation.ts
cowboyd: I saw that you had fixed that, so I did that here too. The difference is that we call the constructor recursively so that the `services` array is rebound as a the next function parameter each time.

More testing sounds like a great idea.

PR #10, opened 2021-03-23
Remove builder syntax

cowboyd:
Motivation
-----------

The builder syntax is getting in the way of rapid innovation. Rather than iterating on the simulation server and quickly adding the scenarios feature, I find that I'm fighting the builder syntax that I was advocating for in the beginning 🤦🏼 The idea was that there would be a nice chaining builder syntax that would help guide you on the right path, but right now it feels like a millstone and this is early days yet.

Approach
----------
This rolls back pretty much anything fancy. The job of a builder object is to emit a data structure, and so this just replaces that to allow simulators to provide the data structure directly.

Before

```js
export const simulator = (behaviors) =>
  behaviors.
    http(app =>
      app.get('/', echo))
    .https(app => app.get('/', echo), 'secure-echo')
```

After

```js
import { createHttpApp } from '@effection/http'; // this will probably live here eventually.

export const simulator = () => ({
  services: {
    echo: {
      protocol: 'http',
      app: createHttpApp().get('/', echo)
    },
    'secure-echo': {
      protocol: 'https',
      app: createHttpApp().get('/', echo)
    }
  }
})
```

This is only slightly more verbose, but it is a hell of a lot easier to implement, and if we want some fancy way to build up those data structures, we can always add them later. But as-is, they are only getting in the way. For now, let's just pass them through transparently.


packages/server/src/simulation.ts
```diff
@@ -62,3 +61,19 @@ export function simulation(definitions: Record<string, Simulator>): Effect<Simul
     }
   };
 }
+
+function selectBehaviors(simulators: Record<string, Simulator>, selected: string[]): Behaviors {
+  return selected.reduce((behaviors, selection) => {
+    assert(simulators[selection] != null, `unknown simulator ${selection}`);
```
dagda1, line 67: small thing and this can be set in eslint to disallow `==` and `!=` and favour the strict `===` and `!=='.

So rather than coalese `null` to `undefined`, is it not better to say `typeof simulators[selection]` or `!!simulators[selection]`?

if you think it reads better then that is fine also.

dagda1 (approved):
Looks good to me and I totally get the motivation.

One small comment.  I am approving so you don't have to do anything if you disagree :).

packages/server/src/simulation.ts
cowboyd: I think it's muscle memory from my CoffeeScript days

PR #11, opened 2021-03-24
Add a person simulator

cowboyd:
Motivation
-----------
We need to do more than just bind services. We have to be able to declare scenarios that arrange the state of the simulation into a specific order.

Approach
----------
This adds the concept of a "scenario" to the simulation and the given() GraphQL API that can be used to add that it to the current simulation.

As the first implementer of this scenario API is the "person" simulator which currently exports a single scenario called "person" which generates the person and inserts it into the store.

The person generation logic is deliberately hokey and we can upgrade it to use faker / fast-check, or some combination later on.

### TODOs and Open Questions

- Currently there is some duplication of the scenario data because we have to actually return the generated data to the graphql caller. I'm not aware of a better way to do this without some sort of referencing mechanism in the atom

### DIFF
--- a/packages/server/src/simulation.ts
+++ b/packages/server/src/simulation.ts
@@ -1,5 +1,5 @@
 import assert from 'assert-ts';
-import { Effect } from './effect';
+import { Effect, map } from './effect';
 import express, { raw } from 'express';
 import { Behaviors, SimulationState, Simulator } from './interfaces';
 import { AddressInfo, createServer } from './http';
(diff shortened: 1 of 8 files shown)


dagda1:
120% agree that the atom is the true path to data replication.

packages/server/src/simulation.ts
```diff
@@ -43,6 +43,30 @@ export function simulation(definitions: Record<string, Simulator>): Effect<Simul
...
+      let { scenarios } = behaviors;
+
+      scope.spawn(map(slice.slice("scenarios"), slice => function*() {
+        try {
+          let { name } = slice.get();
+          let fn = scenarios[name];
+          assert(fn, `unknown scenario ${name}`);
+
+          let data = yield fn(store);
+          slice.update(state => ({
```
dagda1, line 56: definitely nice to have some abstraction over this `update` error.  Something like an 

`type Either<E, A> = Left<E> | Right<A>`

With the 2 alternative paths.

But this is totally fine for now

packages/server/test/person.test.ts
```diff
@@ -0,0 +1,54 @@
...
+    let endpoint = `http://localhost:${port}/graphql`;
+    client = new GraphQLClient(endpoint, { headers: {} });
+  });
+
+  describe('createSimulation()', () => {
+    let simulation: Record<string, any>;
+
+    beforeEach(function*() {
+      let createSimulationMutation = gql`
+mutation CreateSimulation {
```
dagda1, line 24: #15 will cure this.  It is definitely too bulky to write all tests as graphql

dagda1 (approved):
Looks great, really liking the way state changes declaratively make things happen without the endless boilerplate of redux.

PR #16, opened 2021-03-24
Add option to make predictable simulation sequence

cowboyd:
Motivation
-----------

In order to make testing and development easier, we wanted the ability to pass in a uuid to make sure that the created simulation id was stable so that queries we had sitting around would continue to work and sequences of snippets used for testing would not be one-use and obsolete.

Approach
----------

We can accomplish the same goal by just using a predictable sequence of UUIDs. That way, when we run a server, and start a simulation, it always gets the same UUID.

Currently, faker does not support independent instances that each have their own pseudo random number generators so this change in its current form depends on an open PR I have on the faker.js repo making multiple faker instances safe.

We'll need this very soon since each simulation will have its own random seed and its own faker instance.

### DIFF
--- a/packages/server/src/faker.ts
+++ b/packages/server/src/faker.ts
@@ -1,8 +1,10 @@
 // @ts-ignore
 import Faker from 'faker/lib';
+// @ts-ignore
+import locales from 'faker/lib/locales';
 
 export function createFaker(seed: number): any {
-  let faker = new Faker({ locales: require('faker/lib/locales') });
+  let faker = new Faker({ locales });
   faker.seed(seed);
   return faker;
 }
--- a/packages/server/src/faker.ts
+++ b/packages/server/src/faker.ts
@@ -1,0 +1,14 @@
+// @ts-ignore
+import Faker from 'faker/lib';
+
+export function createFaker(seed: number): any {
+  let faker = new Faker({ locales: require('faker/lib/locales') });
+  faker.seed(seed);
+  return faker;
+}
+
+
+export function stableIds(seed: number): () => string {
+  let faker = createFaker(seed);
+  return () => faker.datatype.uuid();
+}
--- a/packages/server/src/schema/context.ts
+++ b/packages/server/src/schema/context.ts
@@ -1,16 +1,15 @@
 import { Slice } from '@effection/atom';
 import { Task } from 'effection';
-import { v4 } from 'uuid';
 import { SimulationState } from '../interfaces';
 
 export class SimulationContext {
-  constructor(private scope: Task, private simulations: Slice<Record<string, SimulationState>>) {}
+  constructor(private scope: Task, private simulations: Slice<Record<string, SimulationState>>, private newid: () => string) {}
 
-  async createSimulation(using: string | string[], uuid?: string): Promise<SimulationState> {
+  async createSimulation(using: string | string[]): Promise<SimulationState> {
     let simulators = ([] as string[]).concat(using);
-    let { scope, simulations } = this;
+    let { scope, simulations, newid } = this;
 
-    let id = uuid || v4();
+    let id = newid();
     let simulation = simulations.slice(id);
     simulation.set({ id, status: 'new', simulators });
 
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -31,6 +31,8 @@
     "cross-fetch": "^3.1.0",
     "effection": "=2.0.0-preview.5",
     "express": "^4.17.1",
+    "faker": "cowboyd/faker.js#one-random-number-generator-per-faker",
+    "@types/faker": "^5.1.7",
     "get-port": "^5.1.1",
     "graphql": "^15.5.0",
     "nexus": "^1.0.0",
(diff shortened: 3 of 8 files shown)


packages/server/src/faker.ts
```diff
@@ -0,0 +1,14 @@
+// @ts-ignore
+import Faker from 'faker/lib';
+
+export function createFaker(seed: number): any {
+  let faker = new Faker({ locales: require('faker/lib/locales') });
```
dagda1, line 5: why do we need to use require here?

packages/server/src/schema/context.ts
```diff
@@ -1,16 +1,15 @@
...
+  constructor(private scope: Task, private simulations: Slice<Record<string, SimulationState>>, private newid: () => string) {}
 
-  async createSimulation(using: string | string[], uuid?: string): Promise<SimulationState> {
+  async createSimulation(using: string | string[]): Promise<SimulationState> {
     let simulators = ([] as string[]).concat(using);
-    let { scope, simulations } = this;
+    let { scope, simulations, newid } = this;
 
-    let id = uuid || v4();
+    let id = newid();
```
dagda1, line 12: does this guarantees the same uuids each time if I do:

```ts
for (const i = o; i < 5 ; i++) {
  console.log(newid());
``

**Review by dagda1 — CHANGES_REQUESTED** [review]
I think it is a good idea to have the uuid for the seed, I looked at using fast check `fc.uuid` but fast-check definitely had issues running outside of its runner.

Do you think you could add a test that would help for documentation also?  Any test at all would be great

File: packages/server/src/schema/context.ts (hunk 2 above)
Comment (cowboyd) [reply packages/server/src/schema/context.ts:12]: Yep. here's a quick test:

```ts
import { createFaker } from './faker';

[1,2,3].forEach(trial => {
  console.log(`trial ${trial}`);
  let faker = createFaker(100);
  [1,2,3,4,5].forEach((i) => {
    console.log(`  ${i}. ${faker.datatype.uuid()}`);
  });
});
```

```
trial 1
  1. 8a4668d2-0212-4a3d-b2d9-1e938281034f
  2. fc221d44-a6bf-4ade-9429-52084ac20997
  3. 96176607-e0fc-408e-a98b-fa2910d318dc
  4. d4140d0f-5ea5-4091-9e5c-2c3a09826991
  5. a523e8f6-9f62-45b3-84b3-d28f0fe95b85
trial 2
  1. 8a4668d2-0212-4a3d-b2d9-1e938281034f
  2. fc221d44-a6bf-4ade-9429-52084ac20997
  3. 96176607-e0fc-408e-a98b-fa2910d318dc
  4. d4140d0f-5ea5-4091-9e5c-2c3a09826991
  5. a523e8f6-9f62-45b3-84b3-d28f0fe95b85
trial 3
  1. 8a4668d2-0212-4a3d-b2d9-1e938281034f
  2. fc221d44-a6bf-4ade-9429-52084ac20997
  3. 96176607-e0fc-408e-a98b-fa2910d318dc
  4. d4140d0f-5ea5-4091-9e5c-2c3a09826991
  5. a523e8f6-9f62-45b3-84b3-d28f0fe95b85
```

packages/server/src/faker.ts
cowboyd: We _may_ want to defer to loading locales depending on the simulation, but best to keep it simple for now. I'm just going to import it.

packages/server/src/schema/context.ts
dagda1: nice!

cowboyd:
@dagda1 I added tests to make sure the ids were stable, and then also refactored the tests a bit to make them a bit easier to setup a server. the `createTestServer` starts a server and then returns a client for it.

dagda1 (approved):
Deterministic data is essential and great to see it there from the start.  Faker is a better fit than fast-check on retrospect.

cowboyd:
> Deterministic data is essential and great to see it there from the start. Faker is a better fit than fast-check on retrospect.

@dagda1 I tried using fast check to start with but while it seemed much more powerful, it also seemed much more geared towards generating edge-cases that were nevertheless valid. For example making sure your system handles an email addres like `aoeunthoeuh+..aatgxvaagg$$@da-oaeu--eogg.11aof.com` which is really handy if you want to stress-test your system, but is not something you would every see IRL. 

We may be able to use it in the future, and we may be able to use the FC integration that you had in the spike in order to make faker based arbitraries.

One thing that kinda stinks right now is that the faker typings are pretty broken. The author has indicated that he would be amenable to accepting typings https://github.com/Marak/faker.js/issues/1023 So perhaps we can propose a setup similar to the one we have in effection v1 where we have typings tests using `dslint`

dagda1:
>>  tried using fast check to start with but while it seemed much more powerful, it also seemed much more geared towards generating edge-cases that were nevertheless valid

@cowboyd  exactly :) it actually tries to break things also.  There are ways to circumvent that but I think it would be hammering fast-check into the solution rather than it being a good fit.

dagda1:
> One thing that kinda stinks right now is that the faker typings are pretty broken. The author has indicated that he would be amenable to accepting typings [Marak/faker.js#1023](https://github.com/Marak/faker.js/issues/1023)

That is interesting, I must take a look

PR #18, opened 2021-03-30
✨Introduce standalone ui application

cowboyd:
Motivation
-----------
We plan to use [graphql-ws][1] as the primary transport for the simulacrum server. There are a lot of good reasons for this, but the most important are:

1. we don't have to implement a protocol ourselves
2. The protocol is cross platform and there are clients for many
different languages
3. As the ecosystem evolves, more tools will form around this protocol.

However, after much digging, it became clear that the graphiql that ships with `express-graphql` is stale and does not support the more recent version of the protocol. There are several issues open on github to upgrade the version of graphiql that ships with, none of which have received attention in a long time. In fact, it appears that the express-graphql project, while not quite abandoned, is not very actively maintained.

Approach
----------

In order to use a modern graphiql, so that we can implement our own fetcher, we have to roll our own app that uses the graphiql react component. Rather than vendoring graphiql, this bundles a complete app with parcel that contains graphiql, and serves that app from the server.

While this is currently in the server package for now, this change is made with the understanding that it will be extracted it into it's own package, `@simulacrum/ui` when we have npm publishing set up. Hopefully in time, the UI can grow to be more than just a GraphiQL instance, but can grow to be a more featureful UI of which one tab is just the graphiql instance.

To that end, it is possible to run the UI, _not_ served by the server. You can currently pass the server url as a parameter to the UI if you want to hack on the UI by itself with the `?server=` query param. In that case you would start your server, and then start the UI separately with a call to:

```
> parcel server ui/index.html --target ui
```

And then navigating to
`http://localhost:1234?server=http://localhost:3000`

[1]: https://github.com/enisdenjo/graphql-ws

### Alternate Designs

I tried a bunch of different takes on this trying to get `express-graphql` to serve a version of graphiql that would work with websockets. If there is an easy way to do this, that perhaps I missed, it would probably have been preferable to shaving this yak.

## Screenshots

This demonstrates running a standalone UI build so that you can hack on it while the server is running

run parcel dev server on port 1234 to connect to simulacrum server on port 3000

Normally however, you would just connect to the simulacrum server and it will serve you its UI


packages/server/ui/index.tsx
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+import { render } from "react-dom";
+
+import { App } from './app';
+
+render(<App server={getServerURL}/>, document.getElementById("root"));
+
+function getServerURL(): URL {
+  let location = new URL(document.location.href);
+  let server = location.searchParams.get('server');
```
dagda1, line 10: we should put this configurable querystring param in the README for people like me who forget

dagda1 (approved):
That is a pity about `graphql-ws`, was there much involved in forking and updating `express-graphql `?  

This will definitely keep things moving which is probably the right thing to do.

I would be inclined to put the `ui` into its own pack now and just bite the bullet.  Things take root the longer they are there.

I'd also have a simple test, just to check the core functionality works.

I'm going to approve so totally up to you.

packages/server/ui/index.tsx
cowboyd: @dagda1 
> That is a pity about `graphql-ws`, was there much involved in forking and updating `express-graphql `?

There was, and in the end, I couldn't get it to work. I think we'll need a UI for this anyway, so I figure this can be the starting point.
 
> I would be inclined to put the `ui` into its own pack now and just bite the bullet. Things take root the longer they are there.

I agree and agonized over this, and believe it or not, the reason I think we should keep it in-package for now is that we are using `npm`, not `yarn`, and we have not yet started publishing our packages to `npmjs.com` yet. Even though `npm` knows to link internal monorepo packages against each other, it still loses its mind on how to do so when there is no public version available 🤦🏻 

Development got a lot easier once we removed the dev dependency on `@simulacrum/client` since it means that you can use the CLI to add and remove deps rather than editting the package.json inline. Anyway, I agree, but didn't want tackling that yak to block this feature.



> 
> I'd also have a simple test, just to check the core functionality works.
> 

👍🏻

dagda1: no problem keeping it in the same package for now.  I just get nervous when a bundler suddenly appears.  Years of webpack have scarred me.

PR #19, opened 2021-03-30
⬆Upgrade Effection

cowboyd:
## Motivation
- The node process apis are now inline with the "resource" / "runnable" API patter that we've been using elsewhere.
- The `once()` method on atom is deprecated

#Approach
Upgrade to use the latest versions of the APIs. No functionality is changed.

## Learning

- https://github.com/thefrontside/effection/pull/269
- https://github.com/thefrontside/effection/pull/285


packages/server/watch.ts
```diff
@@ -50,7 +49,8 @@ function buildAndRun(delay: number):Operation<void> {
       yield executeAndOut('clean');
       yield executeAndOut('generate');
       yield sleep(delay);
-      let server: StdIO = daemon(scope, 'node dist/start.js');
+
+      let server: StdIO = daemon('node dist/start.js').run(scope);
```
dagda1, line 53: is there a `Runnable<T>` interface in effection?

packages/server/src/schema/context.ts
```diff
@@ -15,7 +15,7 @@ export class SimulationContext {
     let simulation = simulations.slice(id);
     simulation.set({ id, status: 'new', simulators, scenarios: {}, store: {} });
 
-    return scope.spawn(simulation.once(settled));
```
dagda1, line 18: I'm kind of sad to see `once` go.  Was the motivation to get people more used to the API?  Could `once` not call `filter(o).expect()`

It is no big deal but slightly curious

dagda1 (approved):

packages/server/watch.ts
cowboyd: Not currently. The idea is that we're not keen to rush into abstractions that can be difficult to unwind, but instead let the formal abstractions brew in the applications that use them. However, when there is a pattern that is useful, such as always making it such that anything that spawns new tasks can be late-bound with a unary function that takes the parent, we want to stick to it. That way, if we do make a formal abstraction around it later, it will be eaiser to migrate towards.

Currently, a nice solution might be to have a method on `Task` called `using()` that would take an object like this, so that you could re-write the above with:

```ts
let server = scope.using(daemon('node dist/start.js'));
//just calls resource.run(self) under the covers, or something like that.
```

packages/server/src/schema/context.ts
cowboyd: I like the once method too. The main thing though is that it's extra API to add to atom that you need to know beyond what is provided by Stream. The thing that's nice now is that if you know stream, you don't need anything else.

What about adding `once()` upstream to `Stream`? That could be nice.

/cc @jnicklas

packages/server/watch.ts
dagda1:
```ts
let server = scope.using(daemon('node dist/start.js'));
//just calls resource.run(self) under the covers, or something like that.
```

I really, really like that.  So simple to write but doing sooo much for the dev.  I'm really excited about effection after having a look at the code and understanding it more.

One thing I was thinking is some sort of convention for disposing of things,....but that is outside the bounds of this PR :).

PR #20, opened 2021-03-31
Break apart graphql context

cowboyd:
## Motivation
\Integrating websockets requires some pretty big changes to the way we execute graphql queries. Specifically, the Effection scope will be different when we do queries over http than when we do queries over websockets. The scope of the http requests will be shared for all requests, but for the subscription we'll want a scope that is per-websocket. Because of this, we won't be able to use a one size fits all graphql context.

## Approach

This breaks apart the context, which was more of a "resolver holder" into a bunch of free standing functions that take the bare minimum of things which they need to run: an effection context, an atom, and an id sequence.

The schema can now just include these free standing functions by passing through the context. It is up to the http and wss layer to choose the appropriate effection scope.

### DIFF
--- a/packages/server/src/schema/operations.ts
+++ b/packages/server/src/schema/operations.ts
@@ -1,0 +1,14 @@
+import { Resolver } from './resolvers';
+import * as resolvers from './resolvers';
+
+export const createSimulation = createResolve(resolvers.createSimulation);
+
+export const given = createResolve(resolvers.given);
+
+function createResolve(resolver: Resolver<unknown, unknown>) {
+  return {
+    resolve(_: any, args: any, context: any): Promise<any> {
+      return resolver.resolve(args, context);
+    }
+  };
+}
--- a/packages/server/src/schema/operations.ts
+++ b/packages/server/src/schema/operations.ts
@@ -1,11 +1,17 @@
 import { Resolver } from './resolvers';
 import * as resolvers from './resolvers';
 
-export const createSimulation = createResolve(resolvers.createSimulation);
+export const createSimulation = uncover(resolvers.createSimulation);
+export const given = uncover(resolvers.given);
 
-export const given = createResolve(resolvers.given);
-
-function createResolve(resolver: Resolver<unknown, unknown>) {
+/**
+ * Nexus types are currently a mess. This lets us take a soundly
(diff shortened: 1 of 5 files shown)


packages/server/src/schema/operations.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const given = createResolve(resolvers.given);
+
+function createResolve(resolver: Resolver<unknown, unknown>) {
+  return {
+    resolve(_: any, args: any, context: any): Promise<any> {
+      return resolver.resolve(args, context);
+    }
+  };
+}
```
cowboyd, line 14: We do this here so that we can have strongly typed resolvers and subscribers, but when they get plugged into the schema they're loosy goosy enough that we don't have to fiddle with nexus typing.

packages/server/src/schema/operations.ts
```diff
@@ -0,0 +1,14 @@
+import { Resolver } from './resolvers';
+import * as resolvers from './resolvers';
+
+export const createSimulation = createResolve(resolvers.createSimulation);
+
+export const given = createResolve(resolvers.given);
+
+function createResolve(resolver: Resolver<unknown, unknown>) {
```
dagda1, line 8: should it not be `createResolver`

packages/server/src/schema/resolvers.ts
```diff
@@ -0,0 +1,48 @@
...
+
+    let id = v4();
+    let scenario = simulation.slice('scenarios').slice(id);
+    scenario.set({ id, status: 'new', name: scenarioName });
+
+    return scope.spawn(scenario.filter(settled).expect());
+  }
+};
+
+function settled<T extends { status: 'new' | 'running' | 'failed' }>(value: T): boolean {
```
dagda1, line 46: Should this not be coming from `ScenarioState`

```ts
function settled(value: ScenarioState['status']): boolean {
```
or create a new type

```ts
type ScenarioStatus = ScenarioState['status']
```

dagda1 (approved):
Looks good!

Just a couple of small comments.

packages/server/src/schema/operations.ts
cowboyd: I renamed it to `uncover()` since the only reason is to strip the type information.

PR #21, opened 2021-04-01
Implement websocket transport

cowboyd:
## Motivation

This is the last PR in the quest to support websockets that will allow us to receive realtime updates from the server regarding its state. The motivation for such is explained at length in other PRS.

## Approach
This adds a websocket transport using `graphql-ws` to implement the protocol which gives us lots of goodies including reconnects, timeouts and connection sharing. It also gives us a nice client that we can use to build our javascript API.


packages/server/src/schema/operations.ts
```diff
@@ -1,8 +1,9 @@
-import { Resolver } from './resolvers';
+import { Resolver, Subscriber } from './resolvers';
 import * as resolvers from './resolvers';
 
 export const createSimulation = uncover(resolvers.createSimulation);
 export const given = uncover(resolvers.given);
+export const state = uncover(resolvers.state);
 
 /**
  * Nexus types are currently a mess. This lets us take a soundly
```
dagda1, line 9: really not warming to nexus at all

dagda1 (commented):
It looks good, do you think you could talk me through it tomorrow or Monday?

cowboyd:
> It looks good, do you think you could talk me through it tomorrow or Monday?

@dagda1 sounds good!

dagda1 (approved):

PR #22, opened 2021-04-01
✨  init covector

jbolda:
## Motivation

Initialize covector to begin publishing the packages.

## Approach

A `covector add` (or similar) command is forthcoming, so to use, add a markdown document in `.changes`.

### TODOs and Open Questions

- [x] We should publish each of the packages manually with just a `package.json` as a `0.0.0`. Otherwise, `npm` has a different flow for first publish.
- [x] Add the npm token as a secret: `NPM_TOKEN`

### DIFF
--- a/.changes/config.json
+++ b/.changes/config.json
@@ -8,19 +8,19 @@
     }
   },
   "packages": {
-    "@simulacrum/client": {
+    "client": {
       "path": "./packages/client",
       "manager": "javascript"
     },
-    "@simulacrum/server": {
+    "server": {
       "path": "./packages/server",
       "manager": "javascript",
-      "dependencies": ["@simulacrum/ui", "@simulacrum/client"]
+      "dependencies": ["ui", "client"]
     },
-    "@simulacrum/ui": {
+    "ui": {
       "path": "./packages/ui",
       "manager": "javascript",
-      "dependencies": ["@simulacrum/client"]
+      "dependencies": ["client"]
     }
   }
 }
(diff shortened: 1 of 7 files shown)


cowboyd (approved):
Awesome. `v0.0.0` of each package has been published to npm!

jbolda:
Oh, I remember last night as my head hit the pillow that we need the `NPM_TOKEN` as well in the secrets. I assume I can just pull that off @frontsidejack from 1password, the publish token?

cowboyd:
> Oh, I remember last night as my head hit the pillow that we need the `NPM_TOKEN` as well in the secrets. I assume I can just pull that off @frontsidejack from 1password, the publish token?

Actually, we have organization secrets for @frontsidejack, so I just extended access to them to both the `simulacrum` and `material-ui-interactors` repositories. There should not be both `FRONTSIDEJACK_GITHUB_TOKEN` and `NPM_TOKEN` in the environment

cowboyd:
It worked! `v0.0.1` of all packages were released as desired.

I did notice something in the output funky though: when it went to check the version of existing releases, it did so without scoping in effect. So, for example, when running `npm view` with simulacrum server, it checked the `server` package which found `v1.0.24` instead of checking `@simulacrum/server` which was at `v0.0.0`

https://github.com/thefrontside/simulacrum/runs/2284928632?check_suite_focus=true#step:5:15

Another thing that I noticed was that the tag created was `server-v0.0.1` instead of `@simulacrum/server-v0.0.1`

jbolda:
The first half of your comment should be addressed by #30, but I didn't notice the tag. Sounds like the same situation. I can PR a fix for that too. 🎉

PR #23, opened 2021-04-01
Fix global clean

cowboyd:
## Motivation

`npm run clean` from the project root will not actually clean everything up. Also, running a build is awkward since we have to `cd` into each package and invoke the build script.

## Approach
Luckily, with npm 7.7.0, we have the capability to fan out a command over all the workspaces by passing the `-ws` flag. This updates the `clean`, `build` and `test` commands to use this functionality so that each project can decide for itself what it means to do these things.

## Learning
- https://github.com/npm/cli/releases/tag/v7.7.0
- https://docs.npmjs.com/cli/v7/using-npm/workspaces#running-commands-in-the-context-of-workspaces

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,16 +3,18 @@
   "version": "0.0.0",
   "description": "Composable simulations for testing, development, and application previews",
   "scripts": {
-    "build": "tsc -b ./tsconfig.packages.json && cd packages/ui && npm run build",
-    "clean": "find . \\( -name node_modules -o -name dir2 -o -path name \\) -prune -false -o -name dist | xargs rm -rf",
-    "watch": "npm run build && \"$(npm bin)/tsc\" -b ./tsconfig.packages.json --watch"
+    "clean": "npm run clean -ws",
+    "test": "npm run test -ws",
+    "build": "npm run build -ws",
+    "build:tsc": "tsc -b ./tsconfig.packages.json",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -4,16 +4,17 @@
   "description": "Composable simulations for testing, development, and application previews",
   "private": true,
   "scripts": {
-    "build": "tsc -b ./tsconfig.packages.json && cd packages/ui && npm run build",
```
dagda1, line 7: this build is that points to the `./tsconfig.packages.json` which builds all typescript packages defined in that file.  We might want to keep that and put in a different script like `build:tsc` or something

dagda1 (changes requested):
That is good to know there is a `-ws` switch

package.json
cowboyd: 👍🏻 Sounds good. Once thing that is nice about the global build script is that because it uses `tsc -b` it's really super fast even though it's invoking three separate builds, the typescript builds use the cache after the first one.

dagda1: >  the typescript builds use the cache after the first one.

Yes, that is what those `*.tsbuildinfo` files are.  I've had things go out of sync a number of times.

`tsc -b ./tsconfig.packages.json --clean` will remove the from the whole solution

dagda1 (approved):

PR #24, opened 2021-04-02
Pass data generation API to scenarios

cowboyd:
Motivation
-----------
Right now, we don't provide any support to our scenarios for generating data. That means that each individual scenario that need to generate data is left completely to its own devices. But given that this is going to be very common, this is something that the simulation server should provide out of the box.

Approach
----------
This creates a single instance of faker per simulation, and passes it into each scenario so that the scenario can use it to generate high-fidelity simulation data.

As a bonus, I was able to actually get rudimentary typing on the `Faker` interface. However, the `@types/faker` module is quite behind the actual faker and so there are a lot of methods on the faker instance that aren't represented in the typings. Given that caveat, it seems to me better than just having it typed as `any`.

### DIFF
--- a/packages/server/src/faker.ts
+++ b/packages/server/src/faker.ts
@@ -1,10 +1,18 @@
 // @ts-ignore
-import Faker from 'faker/lib';
+import FakerInstance from 'faker/lib';
 // @ts-ignore
 import locales from 'faker/lib/locales';
 
-export function createFaker(seed: number): any {
-  let faker = new Faker({ locales });
+// the @types/faker don't account for
+// constructing individual instances of faker so we
+// import the default faker export and use that type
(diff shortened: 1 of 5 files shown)


packages/server/src/faker.ts
```diff
@@ -1,10 +1,18 @@
...
 import locales from 'faker/lib/locales';
 
-export function createFaker(seed: number): any {
-  let faker = new Faker({ locales });
+// the @types/faker don't account for
+// constructing individual instances of faker so we
+// import the default faker export and use that type
+// as the type of all fakers. A nice side-quest would be
+// to add types.d.ts directly to the faker project.
+import faker from 'faker';
```
dagda1, line 11: I keep forgetting to look at this.

dagda1 (approved):

PR #26, opened 2021-04-05
🐛Point to ui app dir, not ui index.html

cowboyd:
Motivation
-----------

In a demo today to the team, the UI broke completely🤦🏻‍♂️ 

With https://github.com/thefrontside/simulacrum/pull/25 we put in a check to see that `dist/index.html` was present, not just the `dist/` directory. I screwed that up by also returning that as the value of `appDir()`. So while it did exist, what the express server needs is the application root directory.

If we can get our bigtest harness up and running for this, it would have caught this.

Approach
----------
Use the `index.html` for the assertion, but return the root directory of the app. Also
- remove the old `ui/index.html` from the `packages/server` app.
- mount the middleware at the root of the server.


dagda1 (approved):
👍

PR #27, opened 2021-04-05
✨@simulacrum/client

cowboyd:
Motivation
----------
One of the super-powers of the simulacrum simulation server is that it has a GraphQL API that can be called from multiple different platforms, such as JavaScript, .Net, Go, Rust, etc... However, it is more convenient not to call through to the low-level graphql API, but instead to use a native language API.

Approach
----------
This introduces the first such language API for JavaScript, and it is built to be an all-purpose JS client for Simulacrum. As such, it does not expose any effection API, but instead uses `Promise`, and `AsyncIterable` externally, even though it uses effection underneath to implement resource disposal and request cancellation.

Rather than have tests for the client itself, this substitutes the client for usage in the server tests. That way, we kill two birds with one stone.

### TODOs and Open Questions

- [x] How do we handle calling unsubscribe from the client actually? If you call `cancel()` on the `AsyncIterator`, that should probably end the iteration? If you call cancel() on a `Promise` does the promise reject?


packages/client/src/index.ts
```diff
@@ -11,10 +21,113 @@ export interface Service {
...
+interface Runnable<T> {
+  run(scope: Task): T;
+}
+
+interface Result<T> {
+  data: T;
+  errors?: GraphQLError[];
+}
+
+export function createClient(serverURL: string, webSocketImpl: unknown = WebSocket): Client {
```
dagda1, line 33: Could you not let typescript infer the type

```ts
export function createClient(serverURL: string, webSocketImpl = WebSocket): Client {
```

dagda1 (approved):
Looks good.  Approving.  1 comment to see if we can get rid of the unknown

packages/client/src/index.ts
cowboyd: We can, but don't we have a linting rule that exported functions should explicitly declare return type?

cowboyd: Oh, you mean the webSocketImpl. The problem is that you can pass anything in there as `graphql-ws` treats it as `unknown`. We could define our own interface perhaps called `WebSocketImpl` but not sure how to go about typing it.

dagda1: totally up to  you if you want to change this, I just feel a bit uneasy about no typing for the `webSocketImpl`

You could do something like this also

```ts
export function createClient<W extends WebSocket = WebSocket>(serverURL: string, webSocketImpl: W): Client {
```

dagda1: Totally up to you if you want to change this.  If there are no other implementations yet, I would always leave just go with `WebSocket` and cross the other bridge when the need arises.

I've approved so this only minor and not necessary.

cowboyd: The problem is that `window.WebSocket` and `ws.WebSocket` do not share a common type, so it does a bunch of weird duck typing https://github.com/enisdenjo/graphql-ws/blob/master/src/client.ts#L273-L291

I'd love to type it too, but not sure the best way.

dagda1: Something like this will work:

```ts
import ws from 'ws';

declare module 'ws' {
  interface WebSocket {
    CLOSED: number;
    CONNECTING: number;
    OPEN: number;
  }
}

export function createClient(serverURL: string, webSocketImpl: ws.WebSocket | WebSocket = WebSocket): Client {
```

It is using declaration merging to add the bits that tsc is complaining about.

We might put the `declare module 'ws'` bit in a `global.d.ts` file or `index.d.ts` so it gets picked  up automatically.

I'm off for a couple of days but I have approved and we could go over it on Friday if I am not understanding the suituation.

cowboyd: What about this?

```ts
export interface WebSocketImpl {
  new(url: string): {
    send(data: string | ArrayBufferLike | Blob | ArrayBufferView): void;
  }
}
```

It matches both `ws.WebSocket` and `window.WebSocket`. basically A constructor function that returns an object with a `send()` method.

dagda1: looks good to me!  More than happy with either.

A public signature with an `unknown` field is a `known` problem :).

PR #43, opened 2021-04-16
✨ Make a proper release

cowboyd:
Motivation
---------
Right now, we're just bundling up the source of Simulacrum and making it downloadable on NPM, but as an executable program, it is a full-featured system written in TypeScript, JavaScript, HTML and CSS. As such, we need to make sure and build all of our sources into their target executables.

Approach
----------
This adds a prepack script to each package that builds its local sources and makes it ready to distribute to npmjs.com. Among other things we:

1. build typescript into javascript
2. add at least a minimal README to every package
3. Fix the author field to not be just cowboyd personally, but all of
FS engineering
4. Only distribute built files, not sources

Finally, I removed the `.graphql` schema generation since we weren't using it anywhere, and all it did was slow down and murk up the build. basically, that stuff is if you want to consume a schema from another package, but all our things consume the schema over HTTP.

> Note:  I left in the aliases to `npm run build` as just doing a prepack. This is because `build` is a little more common nomenclature. However, we can remove these at some point.

### DIFF
--- a/packages/server/src/schema/schema.ts
+++ b/packages/server/src/schema/schema.ts
@@ -2,15 +2,9 @@ import { makeSchema } from 'nexus';
 import path from 'path';
 import { types } from './types';
 
-const parent = path.basename(path.join(__dirname, '..'));
-
 export const schema = makeSchema({
-  shouldGenerateArtifacts: parent !== 'dist',
+  shouldGenerateArtifacts: false,
   types,
-  outputs: {
-    schema: path.join(__dirname, '../__generated__/schema.graphql'),
-    typegen: path.join(__dirname, '../__generated__/schema.types.d.ts'),
-  },
   sourceTypes: {
     modules: [
       {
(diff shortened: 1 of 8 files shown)


packages/server/src/schema/schema.ts
```diff
@@ -2,15 +2,9 @@ import { makeSchema } from 'nexus';
 import path from 'path';
 import { types } from './types';
 
-const parent = path.basename(path.join(__dirname, '..'));
-
 export const schema = makeSchema({
-  shouldGenerateArtifacts: parent !== 'dist',
+  shouldGenerateArtifacts: false,
```
dagda1, line 6: That is interesting, why is always `false`

dagda1 (approved):
approving with 1 question

packages/server/src/schema/schema.ts
cowboyd: We weren't ever using it and so all it did was add to the build time. Now when watching,  restarts are near instantaneous. 

I think this is only useful if you are exporting the schema from the package so that it can be used elsewhere, but we're not doing that, and the graphql endpoint servers the schema anyway. Maybe we can use that to generate TypeScript definitions if we need to.

PR #47, opened 2021-04-16
Debounce watch script

cowboyd:
Motivation
-----------
When I was reviewing the watch script it occured to me that the debounce logic wasn't right. 

The delay needs to go at the beginning of the process, otherwise, we just launching into the middle of the process and not debouncing anything.

## Approach

This optimizes the watch script by having the restart process `sleep(10)` as the very first thing. That way if any file events come in during that time, they'll just cancel the current sleep and sleep again. Once an event has not happened for 10ms, then the process can continue.

### DIFF
--- a/packages/server/watch.ts
+++ b/packages/server/watch.ts
@@ -13,7 +13,7 @@ main(function* (scope: Task) {
       process = scope.spawn(function*() {
         yield sleep(10);
         console.log('rebuilding.....');
-        process = scope.spawn(buildAndRun);
+        yield buildAndRun;
       });
     });
   } finally {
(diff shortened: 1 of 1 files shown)


packages/server/watch.ts
```diff
@@ -1,19 +1,20 @@
...
 
-    yield on(watcher, 'all').forEach(function () {
-      console.log('building.....');
+    yield on(watcher, 'all').forEach(() => {
       process.halt();
-      process = scope.spawn(buildAndRun(500));
+      process = scope.spawn(function*() {
+        yield sleep(10);
+        console.log('rebuilding.....');
+        process = scope.spawn(buildAndRun);
```
cowboyd, line 16:
```suggestion
        yield buildAndRun;
```

dagda1 (approved):
👍

PR #50, opened 2021-05-07
Introduce resources

cowboyd:
## Motivation
Explicitly threading scopes for long running processes is annoying and error prone because you have to a) keep a reference to the containing scope and b) call `spawn` synchronously which can introduce race conditions if the spawned process fails before its first yield point and c) jump through hoops to pass values back and forth between synchronous and asynchronous code

We had mitigated this somewhat by making a `Runnable` interface to normalize all of our long running processes, but it was still annoying for the reasons above and the resulting API was convoluted since the `run()` call returned synchronously and so answers to questions about the `Runnable` were not available at the moment the caller received the runnable.

That's why we had to use deferreds to determine internally when the `http` server was actually listening and had a network address.

https://github.com/thefrontside/simulacrum/blob/459f9d391bb99708c8db558662fab1471771a832/packages/server/src/http.ts#L19-L57

The start up required getting the _address_ of the server to be an operation that waited until the server was running.

```ts
run(function*(scope) {
  let server = createServer().run(scope);
  let { port } = yield server.address();
});
```
To solve this problem, and to make working with never ending tasks, we [introduced resources](https://github.com/thefrontside/effection/pull/309): operations that create long-lived tasks, and return control to the caller only once the resource has been successfully created (or throws an error into the caller if creating the resource fails).

By changing things to use resources and not runnables, not only do we have less code to type and no need to explicitly spawn in the correct scope, we also have the resource initialization being run so that we don't get a reference to the server at all until it is ready and bound to its network address.

The `address` field is now just a regular old property and not an operation because it is guaranteed to be available by the time anybody would want to access it. This allows us to re-write the resource more cleanly https://github.com/thefrontside/simulacrum/blob/998c740d506ff343c5f221f8be4267eafe3af1ed/packages/server/src/http.ts#L17-L46

And it also lets us consume it more intuitively, since we yield to the server creation, and just simply read the address:

```ts
let server = yield createServer();
let address = server.address;
```

## Approach

Replace all `Runnable` instances with resources. This was really straightforward. The only hitch was in running with effection mocha, the resources were not being spawned in the correct scope and not outliving the `beforeEach` block. As a result, we ended up needing to explicitly spawn the resource in `world`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,8 +6,6 @@ jobs:
     docker:
       - image: cimg/node:15.14.0
     steps:
-      - node/install:
-          npm-version: 7.11.2
       - checkout
       - node/install-packages
       - run: npm run build
--- a/package-lock.json
+++ b/package-lock.json
@@ -1669,13 +1669,13 @@
       }
     },
     "node_modules/@effection/atom": {
-      "version": "2.0.0-preview.8",
-      "resolved": "https://registry.npmjs.org/@effection/atom/-/atom-2.0.0-preview.8.tgz",
-      "integrity": "sha512-vRZYg4PsVh6PcniCbG3IlmzkhPuUeI6+cZvsP7D2im8WOkPvHrYQy/Yf18cDwSVw/+Hc4KkCepegsunKLLwJJg==",
+      "version": "2.0.0-preview.10",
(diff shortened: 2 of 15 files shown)


dagda1 (approved):

taras (approved):
👍

PR #52, opened 2021-05-18
Add Volta configuration

cowboyd:
## Motivation

We ran into some issues where the correct version of `npm` was not being used and was causing problems, specifically getting stuck on npm@6

Approach
----------
Volta users (which are many) can circumvent this problem entirely if we have a specification of which point release of node and npm we want to use. This creates a volta config at the root, and then every other package inherits that volta config

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,5 +33,9 @@
   },
   "devDependencies": {
     "covector": "^0.5.3"
+  },
+  "volta": {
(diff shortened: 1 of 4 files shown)


elrickvm (approved):

dagda1 (approved):
I'm still shaken from my volta npm experience earlier

PR #54, opened 2021-05-22
Pass options to simulation when creating

cowboyd:
Motivation
----------

Sometimes you want to determine ahead of time aspects of how a simulation will behave. For example, what if you want a stable port number for your service? This is handy for development if you always want your server to be at the same place. You can make your similulator hard-code a port, or even use an environment variable, but that doesn't scale across running tests in parallel. In that case, what we'd really like to do is pass the port number from _outside_, when we're creating the simulation, by saying "this is the port number to use".

In other words,

```graphql
createSimulation("my-backend", {
  port: 4400
})
```

Approach
----------
This allows for the passage of an arbitrary JSON blob to the simulator that is controlling the simulation. Where as before, a simulator was a no-arg function `() => Behaviors`, it now takes options to become `(options) => Behaviors`. That way, the options can inform how the services and scenarios will behave.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1795,18 +1795,6 @@
         "effection": "^1.0.0"
       }
     },
-    "node_modules/@effection/mocha": {
-      "version": "2.0.0-preview.2-f9e72f1",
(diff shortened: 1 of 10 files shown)


dagda1 (approved):
Looks good to me

PR #55, opened 2021-05-22
Autofix lint and install dependencies with Volta in CI

taras:
## Motivation

When PRs like #54 fail because of a missing semi, every moment spent fixing the semi is a waste of time.
Let computers do stupid work like write semicolons.

## Approach

Try [wearerequired/lint-action](https://github.com/wearerequired/lint-action) which has an `auto_fix` action that fixes lint errors that eslint can fix.

Also: using `volta-cli` to install project dependencies on Lint action.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -8,13 +8,14 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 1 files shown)


dagda1 (approved):
👍

PR #60, opened 2021-05-26
⬆️ Upgrade effection

cowboyd:
## Motivation
effection 2.0.0-preview.14 introduced a `spawn` resource that allows spawns to happen as operations. This is safer because spawning now happens at a yield point, and so if the spawned process fails immediately, we'll never execute the next line of code and willi nstead fail. This should get ride of the dreaded "cannot spawn on a task that is not running"

## Approach
This replaces all of the instances where `spawn` happens inside a generator with the spawn operation.


dagda1 (approved):
👍

PR #73, opened 2021-06-03
upgrade effection in auth0

jbolda:
## Motivation

To make it easier to create a binary, we want to use `ctrlc-windows@2.0.0`. This upgrades to a version of affection which includes that. Also in getting the auth0 tests set up to make certain I didn't break anything in the upgrade, I added a readme so others can more easily get it running on their end.


jbolda:
Failing test is due to the missing cert. I suspect that we will address this before it is merged back into the main branch?

packages/auth0/README.md
```diff
@@ -0,0 +1,27 @@
...
+```
+
+## Contributing
+
+Running the tests requires certs. The easiest setup we have found is by using [mkcert](https://github.com/FiloSottile/mkcert).
+
+Install `mkcert` as noted in their readme, and run the following command in your user folder (depends on OS) at `.simulacrum/certs`.
+
+```shell
+mkcert -key-file localhost-key.pem -cert-file localhost.pem localhost 127.0.0.1
```
dagda1, line 26: The location is also important, I had this in the other simulation server

```md
## SSL

The auth0 simulator needs to run over https.  [mkcert](https://github.com/FiloSottile/mkcert) makes this pretty easy:

```bash
brew install mkcert
brew install nss  # for firefox

mkdir -p ~/.simulatcrum/certs
cd ~/.resideo/certs

mkcert -install   # Created a new local CA at the location returned from `mkcert -CAROOT`
mkcert localhost  # Using the local CA at CAROOT, create a new certificate valid for the following names
```

I could only get it to work correctly with 

```bash
mkcert -install
mkcert localhost
```

but if you can do it one line then that is better. 

The real answer is to script this but it is quite tricky

dagda1 (changes requested):
Looks good to me if you could just please update the README.md

packages/auth0/README.md
cowboyd: What about once we have a CLI having a `doctor`  command that detects incomplete setup? We could also have a `--fix` option that would allow for automated recovery in the event that it can be handled with a script.

dagda1: > What about once we have a CLI having a `doctor` command that detects incomplete setup? We could also have a `--fix` option that would allow for automated recovery in the event that it can be handled with a script.

Sounds like a good idea, this is definitely a bit that can go wrong so we should take responsibility.  Shelling out to mkcert is the easiest thing possibly the best way if it works on windows?

There is this [npm package](https://github.com/Subash/mkcert) but there was a reason I did not go down this path that I can't quite remember.

dagda1: maybe I just got scared :)

jbolda: I definitely trust Filo Sottile much more than that `mkcert` npm package personally. (They were the one who discovered heartbleed.) Does simulacrum pull the localhost key out of the cert store that you need to use `-install`? All I needed to do was create it in that folder that I noted. Those were the steps that I did to get it to work in my box.

dagda1: > All I needed to do was create it in that folder that I noted. Those were the steps that I did to get it to work in my box

When you say got it to work, do you mean when you open a browser the certificate is trusted?  Here is what mine looks like with no CA store:

cert

In order for a cert to be trusted, it needs to be trusted by something which in this case is a local CA.

The magic combination and only thing that ever worked for me was this:

```bash
mkcert -install   # Created a new local CA at the location returned from `mkcert -CAROOT`
mkcert localhost  # Using the local CA at CAROOT, create a new certificate valid for the following names
```

Are you sure you don't already have a local CA?

dagda1: if I run 

```bash
mkcert -install
mkcert localhost
```

Then I see

cert

jbolda: Oh, you have that readme text in `packages/client` already. I didn't realize that. What I added was the minimal amount I needed to run `npm run test` and have them pass in `packages/auth0`. What are you running to get that UI in the screenshots? I presume not having run the login is the reason that I didn't find needing to do `-install`.

Maybe I will just strip out this readme for the moment and we can give that attention later / as I learn how to get everything running.

jbolda:
I will let you merge it @dagda1 when you are ready since it is your branch.

jbolda:
Looks like we are pulling out parts of the branch into separate PRs, so I will open this against main when we get there.

dagda1:
@jbolda apologies for not getting back to you and thanks for your patience.

PR #74, opened 2021-06-09
Middleware

dagda1:
## Motivation

I am using this as a first PR to break up the auth0 branch into smaller chunks.

This PR adds a `use` function to `createHttpApp` that adds express middleware that are operations

## Approach

A `use` function is added to the builder function

```ts
let app = (times: number) => createHttpApp()
  .use(function* (req, res, next) {
    // do stuff


  next();
})
```

An operation is spawned for each middleware element

```ts
for(let middleware of service.app.middleware) {
  app.use(function(...args) {
    scope.spawn(function* () {
      yield middleware(...args);
    });
  });
}
```

### Alternate Designs

You could just use straight up express middleware functions.

### DIFF
--- a/.changes/middleware.md
+++ b/.changes/middleware.md
@@ -1,0 +1,5 @@
+---
+"@simulacrum/server": minor
+---
+
+Add a `use` function to `createHttpApp` that adds express middleware that are operations.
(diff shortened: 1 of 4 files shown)


cowboyd (commented):
This looks good. I wonder what it would look like if the `next` function was also an operation? It feels safer.

cowboyd (approved):

PR #75, opened 2021-06-10
Https

dagda1:
## Motivation

Another chunk of the auth0 branch that provides a means to create HTTPS services.

## Approach

In order for localhost to run under https, there needs to be locally trusted development certificates.

I've made heavy use of [mkcert ](https://github.com/FiloSottile/mkcert) as it is very easy to use.

The code is expecting the certificates to be in a specified location (`~/.simulacrum/certs').

We definitely want to script this if possible although it does require sudo privileges but I think that deserves a PR all on its own.

### Alternate Designs

We want to allow the certificates location to be configurable and I had an interactive `init` command on the other simulator where users could specify a different configuration or accept the defaults and a config file was written.

What do we think is the the best way of running this in CI?  Should we install mkcert as part of the github action?

### DIFF
--- a/.changes/https.md
+++ b/.changes/https.md
@@ -1,0 +1,4 @@
+---
+"@simulacrum/server": minor
+---
+Add the ability to create https services
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,6 +6,7 @@ jobs:
     docker:
       - image: cimg/node:15.14.0
     steps:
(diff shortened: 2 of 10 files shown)


cowboyd (approved):

cowboyd:
@dagda1 I'm 100%👌🏻 on installing `mkcert` for our tests so that we can connect over https

PR #77, opened 2021-06-14
allow normal express middleware to be added as well as operations

dagda1:
## Motivation

The last middleware PR added a means to add middleware functions that were operations:

```ts
reateHttpApp()
    .use(function* () {
      // etc
    })
```    

This makes adding regular 3rd party middleware like [cors](https://github.com/expressjs/cors) difficult.

## Approach

middleware that is either a regular express `RequestHandler` function or an operation can now be added:

```ts
createHttpApp()
    .use(function* () {
      // operation
    })
    .use(function(req, res, next) {
      // regular function
    })
```

### DIFF
--- a/packages/server/src/guards/guards.ts
+++ b/packages/server/src/guards/guards.ts
@@ -1,0 +1,14 @@
+import { RequestHandler } from 'express';
+import type { Middleware } from '../http';
+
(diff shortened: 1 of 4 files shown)


cowboyd (approved):
Looks good 👍🏼

PR #78, opened 2021-06-15
Auth0 package

dagda1:
## Motivation

Another chunk of the `auth0-simulator` branch.  

This PR adds the new `@simulacrum/auth0` package itself and 1 route and 1 test.

## Approach

Add the package and configuration.


packages/server/src/simulators/person.ts
```diff
@@ -12,21 +12,29 @@ export default function(): Behaviors {
...
 
+    let name = params.name ?? faker.name.findName();
+
     // this is the lamest data generation ever :)
-    let attrs = { id, name: params.name || faker.name.findName() };
+    let attrs = {
+      id,
+      name,
+      email: params.email ?? faker.internet.email(name).toLowerCase(),
+      password: params.password ?? faker.internet.password()
```
cowboyd, line 36: I'm ok putting these here for now, but I was thinking that we'd eventually put them in an atom subscription.

dagda1: This is definitely a short hack for this PR.  We need a more general way of creating good fake data that you can override.  We want to do a best guess at faking without overrides first and then leave it open to be overridable if needed.

When I did the original spike, I iterated over the object fields and then and checked if I could get match on the key name before using [this](https://github.com/google/intermock) to generate some more meaningful data.   I had defaults for the field type also.

Anything like that is definitely a PR all of its own and I would definitely want to get buy-in before writing any code.

> But I was thinking that we'd eventually put them in an atom subscription

What would that look like?

cowboyd (approved):

cowboyd:
Yay! 🥳 🎉

The core platform is much better for all the improvements required to accomodate this simulator.

PR #79, opened 2021-06-16
add a static port to simulator options

dagda1:
## Motivation

At the moment, simulator http servers get a random port each time they are created which makes development painful.

## Approach

The `options` parameter of `createSimulation` is a pretty loose type to keep it generic.

This PR allows an optional `port` property to be added to the options hash.  If no port field exists then a random port is assigned.

```ts
export type SimulationOptions = { port?: number } & Omit<Record<string, unknown>, 'port'>;

export interface CreateSimulationParameters {
  simulator: string;
  options?: SimulationOptions;
}

simulation = yield client.createSimulation("static", { port: 3300 });
```

### DIFF
--- a/.changes/port.md
+++ b/.changes/port.md
@@ -1,0 +1,5 @@
+---
+"@simulacrum/server": minor
+---
+
+Allow an optional port field in the options argument of createSimulation.
--- a/packages/server/src/interfaces.ts
+++ b/packages/server/src/interfaces.ts
@@ -41,12 +41,14 @@ export interface ServerState {
   simulations: Record<string, SimulationState>;
 }
 
+export type SimulationOptions = { port?: number } & Omit<Record<string, unknown>, 'port'>;
+
 export type SimulationState =
   {
     id: string;
     status: 'new',
     simulator: string,
-    options: Record<string, unknown>;
+    options: SimulationOptions;
     scenarios: Record<string, ScenarioState>;
     store: StoreState;
   } |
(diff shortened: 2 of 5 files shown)


cowboyd:
Right now ports are allocated at the service within the simulation, so it is possible, for example, to have a single simulator define multiple services within a simulation. In order to make `port` a global simulation option, it sounds like we should refactor to make the services mounted on a path instead of each getting their own port? (This sounds like it might be a good idea for several reasons)

If so, we would need to our service mounting scheme to go from:

```json
[{
  name: 'auth0',
  url: https://localhost:34554
}, {
  name: 'gaweway',
  url: 'https://localhost:3211'
}]
```

to

```json
If so, we would need to our service mounting scheme to go from:

```json
[{
  name: 'auth0',
  url: https://localhost:4400/auth0
}, {
  name: 'gateway',
  url: 'https://localhost:4400/gateway'
}]
```

dagda1:
> it sounds like we should refactor to make the services mounted on a path instead of each getting their own port? (This sounds like it might be a good idea for several reasons)

You make a very good point and I did not see this :).

One problem I hit with the original spike, when I had all simulators etc. running on the same port and just used to middleware to route the calls was to with a common check that happens in things like webworkers.

The iframe code that is rendered does this check

```ts
if (evt.origin != targetOrigin) {
  return;
}
```

I'll need to do a quick check to see that this works if it is mounted on a URL like `https://localhost:4400/auth0`.

dagda1:
closed in favour of #80

PR #80, opened 2021-06-18
A static port can be assigned to a service from the control API.

dagda1:
## Motivation

Following on from #79, this PR tackles the problem of assigning a static port to a service.

## Approach

The `createSimulation` options argument now takes a `services` field , e.g.

```ts
const simulation = yield client.createSimulation("echo", { options: {
  times: 3
}, services: {
  echo: {
    port: 3300
  }
} });
```

### DIFF
--- a/.changes/port.md
+++ b/.changes/port.md
@@ -2,4 +2,4 @@
 "@simulacrum/server": minor
 ---
 
-Allow an optional port field in the options argument of createSimulation.
+a `services` field added to the `createSimulation` options argument where a port can be assigned.
--- a/.changes/port.md
+++ b/.changes/port.md
@@ -1,0 +1,5 @@
+---
+"@simulacrum/server": minor
+---
+
+Allow an optional port field in the options argument of createSimulation.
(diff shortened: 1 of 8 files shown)


cowboyd (commented):
Why do options and config need to be a separate object?

dagda1:
> Why do options and config need to be a separate object?

Because they are 2 separate things, options are specific to the current object and config would be generic to all simulators.

I don't have any objection to them being all in 1 object but I guess there is the chance of name collisions.

cowboyd:
Hmm.....

In that case, what about making "options" a key in the config?

```js
await client.createSimulation("auth0", { options: {}, services: { auth0: { port: 4400 } } });
```

I think that `options` will be pretty rare, and it lets us get rid of the  top-level`config` key.

dagda1:
> ```js
> await client.createSimulation("auth0", { options: {}, services: { auth0: { port: 4400 } } });
> ```

I definitely think that this is better.

cowboyd (approved):

PR #81, opened 2021-06-29
Auth0 authorize

dagda1:
## Motivation

Add the initial OpenID `/authorize` endpoint.

Initially, the code is dealing with the OpenID authorization code flow `response_type=code`

`/authorize` is the first OpenID endpoint that is called and if the user has not logged in then the server redirects to a `/login` endpoint to ask for credentials.

## Approach

A `/authorize` handler and a `/login` stub handler are added to the simulator in the `@simulacrum/auth0` package.

### TODOs and Open Questions

- [x] The simulator needs to know the actual port of the express server it is running on.

The auth0 simulator needs to know the full URL of the express server that the service is listening on.  The express handlers need to know the domain for the auth0 configuration and this would require knowing the port unless it is statically assigned.

The problem is that the `simulator` function is called before the `createServer` function which is too late to know the port.


cowboyd (commented):
Let's see if we can get rid of nock, and just create service in our tests using simulacrum.

dagda1:
> Let's see if we can get rid of nock, and just create service in our tests using simulacrum.

nock is a gonner

packages/auth0/test/auth0.test.ts
```diff
@@ -4,14 +4,38 @@ import { createTestServer, Client, Simulation } from './helpers';
...
+                  res.status(200).send('ok');
+                })
+              },
+
+            },
+            scenarios: {
+              *person(store, faker) {
+                return yield person(store, faker);
+              }
+            }
```
cowboyd, line 35: You could probably write this as:

```ts
scenarios: { person }
```

dagda1: > You could probably write this as:
> 
> ```ts
> scenarios: { person }
> ```

done

cowboyd (approved):

PR #84, opened 2021-07-06
upgrade effection to v2 beta

jbolda:
## Motivation

The POC around the node binaries requires an updated version of effection which depends on `ctrlc-windows@>=2.0.0`. Upgrading so we can begin to merge in that work.

## Approach

Updated all effection related deps.

### TODOs and Open Questions

- [x] A few of the package scripts used env vars specified in a way that was not compatible on windows. I wrote these into the scripts directly as they seem to operate as defaults. Is this a fair approach here? 
- [x] The `server` uses `NODE_EXTRA_CA_CERTS` in tests. It relies on the env var specified in a way that doesn't work on windows. Unfortunately, that variable is read once when node starts so there isn't a great way to patch it. If I set it ahead of time, I also cannot use `export NODE_EXTRA_CA_CERTS=\"$(mkcert -CAROOT)/rootCA.pem\"` as it returns the incorrect slashes. Node reads this location using posix seperators. It will only work if I use `export NODE_EXTRA_CA_CERTS=~/AppData/Local/mkcert/rootCA.pem` or `export NODE_EXTRA_CA_CERTS=C:/Users/Jacob/AppData/Local/mkcert/rootCA.pem`. I don't know a better way to deal with this.
- [ ] I had to convert from `Deferred` to `Future` in this upgrade. I based these changes on [the changes we made in the effection tests](https://github.com/thefrontside/effection/pull/356), but I would appreciate a second set of eyes here.


dagda1:
could you not use cross-env for the environment variable?  we probably should be using it anyway

jbolda:
> could you not use cross-env for the environment variable? we probably should be using it anyway

@dagda1 Which environment variables would we use it for? The `PORT` and such just seem like defaults, so it felt fine including it within the scripts. `cross-env` relies on `cross-spawn`, and uses a child_process to launch the script. It didn't seem worthwhile to add another node process in the mix for something that feels more like a default.

For `NODE_EXTRA_CA_CERTS`, I didn't think it worked because `cross-env` starts the node process and doesn't have that env var available yet. If it is just running a shell "script" though which starts another node process, maybe it would work? It could I had my cert misconfigured when I tried. Worth trying again if you also think it is worthwhile to add.

If cross-env works for `NODE_EXTRA_CA_CERTS`, I am still a bit inclined to leave the `PORT` and such as defaults rather then embedding it within the script. Thoughts?

jbolda:
Hmm, seems like node dice:

```shell
> @simulacrum/server@0.2.0 test
> cross-env NODE_EXTRA_CA_CERTS="$(mkcert -CAROOT)/rootCA.pem" mocha -r ts-node/register --timeout 10000 test/**/*.test.ts

Warning: Ignoring extra certs from `$(mkcert -CAROOT)/rootCA.pem`, load failed: error:02001003:system library:fopen:No such process
```

jbolda:
Chatting with @dagda1. I think we deal with `NODE_EXTRA_CA_CERTS` in a separate PR, and not add `cross-env` at the moment.

dagda1:
> Chatting with @dagda1. I think we deal with `NODE_EXTRA_CA_CERTS` in a separate PR, and not add `cross-env` at the moment.

totally agree

packages/server/src/start.ts
```diff
@@ -1,4 +1,4 @@
-import { main } from '@effection/node';
+import { main } from '@effection/main';
```
cowboyd, line 1:
```suggestion
import { main } from 'effection';
```

packages/server/watch.ts
```diff
@@ -1,8 +1,12 @@
 import { Operation, Stream, Task, on, sleep, spawn } from 'effection';
-import { main, exec, daemon, StdIO } from '@effection/node';
+import { main } from '@effection/main';
```
cowboyd, line 2:
```suggestion
import { main } from 'effection';
```

cowboyd:
Some of these suggestions would result in de-optimized imports, so probably best not to merge them directly from github

jbolda:
> Some of these suggestions would result in de-optimized imports, so probably best not to merge them directly from github

Ah, didn't know that. Updated!

cowboyd (approved):

PR #85, opened 2021-07-07
add /oauth/token endpoint

dagda1:
## Motivation

This PR adds the `/oauth/token` endpoint to the auth0 simulator that actually issues the token.

## Approach

The handler checks that a person scenario exists in the store with the specified username

```ts

let user = personQuery(([, person]) => {
  assert(!!person.email, `no email defined on person scenario`);

  return person.email.toLowerCase() === username.toLowerCase();
});
```

Another challenge was to ensure that the `nonce` field that is created in the original `/authorize` call is the same `nonce` that ends up in the token that is returned from `/oauth/token`:

```ts
let idTokenData = {
  // other fields
  nonce,
};
```


cowboyd (approved):
I won't pretend to understand the cryptographic aspect of it, but everything else looks groovy.

PR #88, opened 2021-07-12
add atom dependency to auth0

dagda1:
## Motivation

Building `v0` locally is failing due to the missing `@effection/atom` dependency in `@simulacrum/auth0`.

## Approach

Add the dependency to `@simulacrum/auth0`.

### DIFF
--- a/packages/auth0/package.json
+++ b/packages/auth0/package.json
@@ -29,6 +29,7 @@
   },
   "homepage": "https://github.com/thefrontside/simulacrum#readme",
   "dependencies": {
+    "@effection/atom": "^2.0.0-beta.3",
     "@effection/process": "^2.0.0-beta.3",
     "@simulacrum/server": "0.2.0",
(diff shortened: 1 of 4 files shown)


packages/auth0/package.json
```diff
@@ -29,6 +29,7 @@
   },
   "homepage": "https://github.com/thefrontside/simulacrum#readme",
   "dependencies": {
+    "@effection/atom": "^2.0.0-beta.3",
```
cowboyd, line 32: Should this be a dependency, or a dev dependency? Are we importing anything from `@effection/atom`?

dagda1: actually, it was only importing the `Slice<T>` type so I've updated the imports to:

```ts
import type { Slice } from '@effection/atom';
```

And moved it into `devDependencies`

cowboyd (approved):

PR #89, opened 2021-07-14
Rules

dagda1:
## Motivation

[Auth0 rules](https://auth0.com/docs/rules) can be added for numerous [different scenarios](https://auth0.com/docs/rules/use-cases) and are notoriously difficult to test.

This PR adds a basic way of getting the rules code under test.

## Approach

A `rulesDirectory` field has been added to the `auth0` options:

```
mutation {
 createSimulation(simulator: "auth0", 
   options: { 
      options:{
        rulesDirectory: "test/rules"
```

Any rules code files in this directory will be executed while responding to an `/oauth/token` request.

- This assumes the rules code is on disk.  Auth0 rules can be added through the web wysiwyg.  There is nothing in this PR that retrieves the code from auth0.
- The `ruleDirectory` code needs to be the directory of the compiled output and not typescript or code that needs to go through a compiler.

### DIFF
--- a/.changes/rules.md
+++ b/.changes/rules.md
@@ -1,0 +1,5 @@
+---
+"@simulacrum/auth0": minor
+---
+
+Add the initial rules-runner code
--- a/package-lock.json
+++ b/package-lock.json
@@ -2131,6 +2131,12 @@
       "integrity": "sha512-hsPGC/U/0xe/WghMeSgyFsq9nNPfA5oY1Il2AaeAJcu/vTm4Bv8o9ev4eAgxcA61i5WWp72amN20XVyxWwM5aQ==",
       "dev": true
     },
+    "node_modules/@types/auth0": {
+      "version": "2.33.4",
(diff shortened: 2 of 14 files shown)


taras (changes requested):
it looks like a good starting point to show how it’ll work. we’ll need to see how correct it is in practice. can you add a small section to a README for this package describing this functionality and current implementation so it doesn’t get lost. it doesn’t need to be a lot because it’ll need to be covered in documentation but at least if someone is curious about rules they can read it. 

after that i think we can merge it.

dagda1:
> it doesn’t need to be a lot because it’ll need to be covered in documentation but at least if someone is curious about rules they can read it.
> 
> after that i think we can merge it.

I was going to update the README for the auth0 simulator on a separate PR when all the code is in.

There is one small PR after this for `/logout`.

taras:
> I was going to update the README for the auth0 simulator on a separate PR 

it’s fine if that’s easier for you. i just want to make sure we don’t forget.

taras (approved):
👍 with one note that we need to remember to document this in the README

PR #90, opened 2021-07-14
Logout

dagda1:
## Motivation

Implement the `/v2/logout` route to clear the users credentials.

## Approach

Clear the session and redirect the user to the `returnTo` field of the auth0 config.

### DIFF
--- a/.changes/logout.md
+++ b/.changes/logout.md
@@ -1,0 +1,5 @@
+---
+"@simulacrum/auth0": minor
+---
+
+Add `/v2/logout` and remaining pieces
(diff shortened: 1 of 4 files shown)


taras (commented):
conceptually, without knowing how this works in the simulator, i would expect logout to clear the session in the simulator state. are we storing in the atom users who have sessions? if so, do we need to clear that state when logout is called?

dagda1:
> conceptually, without knowing how this works in the simulator, i would expect logout to clear the session in the simulator state. are we taking in the atom users who have sessions? if so, do we need to clear that state when logout is called?

You would not do that in this case because it would remove any scenarios that are created.  You would expect the same user to log back in without re-running the scnearios.

taras (approved):

PR #92, opened 2021-07-14
clarifications

jbolda:
## Motivation

I walked through the readme for the auth0 simulator and did all the steps. There are a few places that I added additional clarification to improve the flow.


dagda1 (commented):
Many thanks for this.  I meant to ask someone to sanity test this.

packages/auth0/README.md
```diff
@@ -1,31 +1,34 @@
...
-  - [code](#code)
-- [Configuration](#configuration)
-  - [options](#options)
-  - [services](#services)
-- [Rules](#rules)
-- [Endpoints](#endpoints)
 
-Please read the [main README](../../README) for more background on simulacrum.
+- [Auth0 simulator](#auth0-simulator)
+  - [Table of Contents](#table-of-contents)
```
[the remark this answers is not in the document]
jbolda: I have a markdown formatter, and ran it on the TOC. Looks like it did it for the whole document. It is a bit wonky that it includes itself and the main heading. I will remove that.

packages/auth0/README.md
```diff
@@ -191,14 +202,12 @@ If we want to run these rules files then we would add the `rulesDirectory` field
 ```ts
 let simulation = yield client.createSimulation(url, {
   options: {
```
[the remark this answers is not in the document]
jbolda: Oh sure enough 👍

packages/auth0/README.md
```diff
@@ -71,58 +74,63 @@ Use the values returned from the query to update your configuration in the clien
...
     port,
-    simulators: { auth0 }
+    simulators: { auth0 },
   });
 
   let url = `http://localhost:${server.address.port}`;
 
   console.log(`simulation server running at ${url}`);
 
-  let client = yield createClient(url);
```
jbolda, line 111: @dagda1 I removed the `yield` here. It was throwing a loop error in effection. Just noting in case this shows up somewhere else that I didn't notice.

packages/auth0/README.md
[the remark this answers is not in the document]
dagda1: turns out you are totally correct, it is only the double options in the graphql.

packages/auth0/README.md
cowboyd: `createClient` is indeed a synchronous function. Also, might be worth showing usage from `async/await` since client is designed to hide the fact that it uses effection internally.

jbolda: Is `createSimulationServer` designed that way as well? Do we still need to wrap it with effection `main` if we are using `async`/`await`? I can't seem to get the server to start using `async`/`await`.

cowboyd: @jbolda No, `createSimulationServer` must be used as an Effection operation. The client on the other hand, will be running inside all kinds of enviroments; mostly cypress, jest, mocha test suites, etc... and so for the client only we opted for making it just simple `async/await` and not assume an effection environment.

The simplest integration is to use `main()` or `run()` if you need to and know what you're doing.

jbolda: @cowboyd I added an additional example that includes async/await. Let me know what you think.

dagda1 (approved):

PR #94, opened 2021-07-15
add console.error in catch handlers

dagda1:
## Motivation

Catch handlers in `simulation.ts`  were not getting logged to the console.

## Approach

add `console.error` to catch handlers.

### DIFF
--- a/packages/server/src/effects/logging.ts
+++ b/packages/server/src/effects/logging.ts
@@ -1,21 +1,26 @@
-import type { SimulationState } from 'src/interfaces';
-import { Effect } from '../effect';
+import type { ServerState } from 'src/interfaces';
+import { map } from '../effect';
 import { assert } from 'assert-ts';
-import { spawn, Task } from 'effection';
+import { Operation, spawn, Task } from 'effection';
+import { Slice } from '@effection/atom';
 
-export const loggingEffect = (): Effect<SimulationState> => slice => function*(scope) {
-  let task: Task = yield spawn();
+export function createLogger(slice: Slice<ServerState>): Operation<void> {
+  return {
+    name: 'logger',
+    *init(scope) {
+      let task: Task = yield spawn();
 
-  yield slice.slice('debug').forEach(function*(shouldLogErrors) {
-    yield task.halt();
-    if (shouldLogErrors) {
-      task = yield spawn(function* () {
-        yield slice.filter(({ status }) => status === 'failed').forEach(function *(state) {
-          assert(state.status === 'failed');
-
-          console.error(state.error);
-        });
-      }).within(scope);
+      yield spawn(slice.slice("debug").forEach(function*(shouldLogErrors) {
+        yield task.halt();
+        if (shouldLogErrors) {
+          task = yield spawn(map(slice.slice("simulations"), function*(simulation) {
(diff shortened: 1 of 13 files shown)


jbolda (changes requested):
I would maybe add some text to this change file. I don't know that covector will handle this nicely. We should probably have it throw an error for this situation. I don't know that we want to allow no change info?

cowboyd:
I don't think this is what we want, since it will be normal that scenarios and services fail. The errors will be in the atom state and should be handled from there. What we could do is, take have an optional effect that sees an error for the first time in the atom and logs it to the console. That way, you could turn this effect on and off from the command line.

dagda1:
> That way, you could turn this effect on and off from the command line.

I did not realise it was getting logged to the state.  Having logging configurable sounds like a better idea

cowboyd:
> Are you happy with the approach?

I think we agreed in chat that we'd make it server wide, not a simulation option?

another thing that occurned to me which is super fancy, but not that much more difficult to implement is that we could make the debug a field in the atom (initial state set by a server option), and start the effect whenever it is true, and halt the effect whenever it is false. That would mean we could turn logging on and off of a running server.

dagda1:
> start the effect whenever it is true, and halt the effect whenever it is false. That would mean we could turn logging on and off of a running server.

That is nice!

packages/server/src/effects/logging.ts
```diff
@@ -0,0 +1,23 @@
...
+      yield task.halt();
+      task = yield spawn(function* () {
+        yield slice.filter(({ status }) => status === 'failed').forEach(function *(state) {
+          assert(state.status === 'failed');
+
+          console.error(state.error);
+        });
+      }).within(scope);
+    } else if (!shouldLogErrors) {
+      task.halt();
```
cowboyd, line 20: Since we halt the task in both cases, it looks like we can just do that at the top once, and only have a single branch if for the positive case.

cowboyd (commented):
This implementation looks very elegant (too bad we don't have a mutation to enable logging ☹️ )

I thought we were going to make it a server option and not a simulation option though?

packages/server/src/effects/logging.ts
dagda1: 👍

dagda1:
> I thought we were going to make it a server option and not a simulation option though?

hmmm, the only place that could be set is in either `createClient` or in the yet to be created `enableLogging` mutation?

packages/server/src/effects/logging.ts
```diff
@@ -0,0 +1,21 @@
...
+import { Effect } from '../effect';
+import { assert } from 'assert-ts';
+import { spawn, Task } from 'effection';
+
+export const loggingEffect = (): Effect<SimulationState> => slice => function*(scope) {
+  let task: Task = yield spawn();
+
+  yield slice.slice('debug').forEach(function*(shouldLogErrors) {
+    if (shouldLogErrors) {
+      yield task.halt();
```
cowboyd, line 11: I think we want to halt in all cases.

```suggestion
    yield task.halt();
    if (shouldLogErrors) {
```

cowboyd (approved):

jbolda (approved):

PR #97, opened 2021-07-16
feat: auth0 service examples

jbolda:
## Motivation

We want to enable users to quickly try out a simulacrum server with the auth0 service. This sets up two examples using nextjs to enable it.

## Approach

This has two examples: one using auth0-react and one using nextjs-auth0.

The former appears to work as expected, and I have added a CSB link. I don't know that it will work until we run our next publish though. The latter appears to have an issue with `socket hung up`. It seems that we are just not handling that auth use case (cookies?), and that can be enabled in a separate PR. I left it in this PR with a note that it is in process in the readme. _Note it was already fixed._

I also hooked it up to covector so it will get an updated changelog which makes it easier to track what we changed over time.

### Alternate Designs

Regarding CodeSandbox, we could also set up the CSB bot to build an updated sandbox on PRs. That may be worth exploring, but I think we have other priorities first before that makes sense to realize.

### TODOs and Open Questions

- Is it OK to commit the example that currently isn't working? This is a new enough library that I am not terribly concerned about it personally.


jbolda:
Sort of an interesting situation, but since the ESlint action ran and made the last commit, only CircleCI ran. By default, actions can't trigger other actions so even though covector successfully ran in my last commit, it didn't run on the last commit. Since it is required, the status check continues to wait for it.

jbolda:
With that last push from @dagda1, I think this is ready to go. It is rather feature slim, but we can iterate from here.

dagda1 (approved):
👍

PR #100, opened 2021-07-27
pass state to auth0-js and fix trailing for openid endpoints

dagda1:
## Motivation

#97 exposed a couple of subtle bugs.

## Approach

There were a couple of glitches fixed in this PR

1.  The same state field was not getting passed throughout the redirects

2. The `/.well-known/openid-configuration` endpoint was not adhering to the required forward slash of the URLs.  The `iss` field of the jwt must have a forward slash.  The json object returned from `/.well-known/openid-configuration` has an `issuer` field which clients use to validate the `iss`.  Our `issuer` field did not have a forward slash.

### DIFF
--- a/.changes/state-iss.md
+++ b/.changes/state-iss.md
@@ -1,0 +1,5 @@
+---
+"@simulacrum/auth0": minor
+---
+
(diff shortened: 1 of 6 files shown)


taras (approved):
👍

PR #113, opened 2021-08-04
add bin script to auth0-simulator so it can be started via npx

dagda1:


taras (changes requested):
I love this idea. I was thinking yesterday that it would be nice for people to be able to start auth0 simulator by itself without worrying about simulacrum. This PR accomplishes just that. 

I got confused for a minute because the changes the the README updated the name of the package, but not the rest of the content which needs to be updated to match a different workflow. 

auth0-simulator starts the simulator by calling the GraphQL mutation so technically we can skip all of that. We need to make that clearer. I'll suggest a different structure in comments.

taras:
```md
## Table of Contents

- [Auth0 simulator](#auth0-simulator)
  - [Table of Contents](#table-of-contents)
  - [Quick Start](#quick-start)
    - [Standalone](#standalone)
    - [Programmatically](#programmatically)
  - [Configuration](#configuration)
    - [Options](#options)
    - [Services](#services)
  - [Rules](#rules)
  - [Endpoints](#endpoints)

Please read the [main README](../../README.md) for more background on simulacrum.

The auth0 simulator has been initially written to mimic the responses of a real auth0 server that is called from auth0 client libraries like [auth0/react](https://auth0.com/docs/quickstart/spa/react/01-login) and [auth0-spa-js](https://github.com/auth0/auth0-spa-js) that use the OpenID [authorization code flow](https://developer.okta.com/docs/concepts/oauth-openid/).

If this does not meet your needs then please create a GitHub issue to start a conversation about adding new OpenID flows.

## Quick Start

### Standalone

The easiest way to run the Auth0 Simulator is with `npx auth0-simulator`. 

[Interally](src/start.ts), it will make an API call to Simulacrum to start the simulator.

- ### GraphQL
+ ### Programatically

This is how you'd integrate the simulator into your test harness because it'll allow you to start and restart the server from a helper in your tests.  
```

dagda1:
> auth0-simulator starts the simulator by calling the GraphQL mutation so technically we can skip all of that. We need to make that clearer. I'll suggest a different structure in comments.

It only starts the simulacrum sever actually, `createSimulation` is not called and I don't think you want to if you think about running from cypress tests.

I'm definitely open to other opinions on this one though of a better way.  It would be great to not even know about graphql as that is a barrier for entry.

taras:
@dagda1 right, I see what you mean. This opens a bit of a can of worms. I'm afraid that without some additional functionality, this might end up more confusing.

It might be confusing because, without additional feedback in their CLI command, they might think that the simulation server is auth0. We should wait for others to give their opinions, but if we do add more functionality to this CLI then we might want something like this.

```
$ npx auth0-simulator
Started Simulacrum simulation server on HTTP://locahost:${port}. 
GraphiQL interface is running on HTTP://locahost:${port}/graphql.

To start auth0 simulator send the following mutation to GraphQL server.

mutation CreateSimulation {
 createSimulation(simulator: "auth0",
  options: {
    options:{
      audience: "[your audience]",
      scope: "[your scope]",
      clientId: "[your client-id]"
    },
    services:{
      auth0:{
        port: 4400
      }
    }
  }) {
    id
    status
    services {
      url
      name
    }
  }
}
```

To start auth0-simulator while starting the CLI,

```
$ npx auth0-simulator start
Started Simulacrum simulation server on HTTP://locahost:4000. 
GraphiQL interface is running on HTTP://locahost:4000/graphql.

auth0-simulator was started on https://localhost:1234
```

taras (approved):
🔥

cowboyd (approved):

PR #114, opened 2021-08-04
Add @simulacrum/cypress-auth0 package

dagda1:
## Motivation

Add a `@simulacrum/cypress-auth0` package that contains `logIn` and `logOut` commands

## Approach

Because we can start the simulacrum server with `npx ` when #113 lands the cypress commands feel like they should be in their own separate package.

The location and naming of things are open for debate but hopefully, the package itself is relatively simple.

## TODO

- [x] Make `createSimulation` idempotent, i.e. creating a simulation on the same port should not crash the whole simulation server.
- [x] Fix logout, currently returns 500 and also make it idempotent so it does not fail if the scenario is not logged in.
- [ ] enable error logging from simulacrum
- [x] Optional fix to provide a better way of maintaining state and remove the ugly `ClientMap`


taras (commented):
This is awesome. I'll let Charles and Jacob do a more thorough review.

cowboyd (commented):
Can we move as much as we can into `before` and `beforeEach` hooks?

examples/nextjs-with-auth0-react/package.json
```diff
@@ -21,7 +21,10 @@
     "react-dom": "17.0.2"
   },
   "devDependencies": {
+<<<<<<< HEAD
     "@simulacrum/auth0-simulator": "0.2.2",
+=======
+>>>>>>> 9d79d40... update config.json
```
taras, line 27: I don't think this will work.

examples/nextjs-with-nextjs-auth0/package.json
```diff
@@ -21,7 +21,10 @@
     "react-dom": "17.0.2"
   },
   "devDependencies": {
+<<<<<<< HEAD
     "@simulacrum/auth0-simulator": "0.2.2",
+=======
+>>>>>>> 9d79d40... update config.json
```
taras, line 27: Wat dis?

taras (commented):
Other then the merge problems 👍

cowboyd (approved):
:shipit:

examples/nextjs-with-nextjs-auth0/package.json
dagda1: oh dear, how sloppy.  Good spot.

.changes/cypress.md
```diff
@@ -1,5 +1,5 @@
 ---
-"@simulacrum/auth0": minor
+"@simulacrum/auth0-simulator": minor
 "@simulacrum/cypress-auth0": minor
```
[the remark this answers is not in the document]
cowboyd: Is there a reason we shouldn't?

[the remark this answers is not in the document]
dagda1: Oh...I missed these.

jbolda (commented):
I also went through and tried to get this working in the example branch, but was having issues with the wrap timing out. I am tempted to say it is a misconfiguration on my end though, and I think maybe we touch on these last things and merge it in. I can rebase my examples and try again.

@dagda1 need me to finish these things up? Let me know thoughts on these I can take make any adjustments if need be.

Also, side note, I don't see any destroying of simulations here, but I don't think we need to worry with the hash; they all get destroyed when the process finishes, right? Maybe file it under an "address in the future" consideration.

integrations/cypress/cypress.example.env.json
```diff
@@ -0,0 +1,7 @@
+{
+  "audience": "https://thefrontside.auth0.com/api/v1/",
+  "domain": "localhost:4400",
+  "client_id": "YOUR_AUTH0_CLIENT_ID",
```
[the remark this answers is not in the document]
dagda1: the reason I did this is that the auth0 config uses `client_id` and it is a source of bugs (I've done it) by passing in `clientId` and not `client_id`.  I think I would rename ours to `client_id` and be done with it.

integrations/cypress/cypress/integration/login.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+        .then(() => {
+           cy.visit('/');
+
+           cy.contains('Log out');
+        })
+        .logout();
+    });
+  });
+
+  describe('', () => {
```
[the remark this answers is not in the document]
dagda1: oh dear, my bad.  This block was to test using member variables instead of chaining

integrations/cypress/src/cypress.ts
```diff
@@ -0,0 +1,102 @@
...
+      out<S = unknown>(msg: string): Chainable<S>
+    }
+  }
+}
+
+const ClientPort = process.env.PORT || 4000;
+
+const ClientMap: Map<string, Client> = new Map();
+
+Cypress.Commands.add('out', (msg: string) => {
```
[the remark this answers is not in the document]
dagda1: kill it with fire

integrations/cypress/src/cypress.ts
```diff
@@ -0,0 +1,102 @@
...
+  if(ClientMap.has(Cypress.spec.name)) {
+    client = ClientMap.get(Cypress.spec.name);
+  } else {
+    client = createClient(`http://localhost:${ClientPort}`);
+    ClientMap.set(Cypress.spec.name, client);
+  }
+
+  assert(typeof client !== 'undefined', 'no client created in createSimulation');
+
+  let { domain, client_id, ...auth0Options } = options;
```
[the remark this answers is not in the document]
dagda1: I think better validation needs to happen in the auth0 simulator and not here.  Errors should be thrown from there.  That is probably a new PR.  We don't want lots of similar validation in multiple places

integrations/cypress/cypress.example.env.json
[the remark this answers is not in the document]
jbolda: https://github.com/thefrontside/simulacrum/issues/124

integrations/cypress/README.md
```diff
@@ -93,3 +93,11 @@ after(async () => {
   cy.logout();
 });
 ```
+
+You may also log out _before_ each test to allow one to inspect the state after a test run.
+
+```js
+beforeEach(async () => {
+  cy.logout();
```
dagda1, line 101: Does this work?  I think this might throw an exception as it currently stands

jbolda: Hmm, good point, I don't think it will. Let me throw in a test to confirm. Speaking of which, I don't think we are running the tests for this in CI. I can add it to the workflow.

dagda1: I think calling `cy.logout()` should be idempotent or not blow up if you are not logged in.

We should possibly leave the README and fix `cy.logout()`

We cold pair on this later if you have time.

jbolda: I can take a stab first. We can pair tomorrow (my) morning if I am stuck if that sounds good with you.

dagda1: go for it man!  All yours

packages/server/src/schema/resolvers.ts
```diff
@@ -20,12 +21,19 @@ export interface CreateSimulationParameters {
...
-    let id = newid();
+    let id = options.key ?? newid();
+
     let simulation = atom.slice("simulations", id);
 
+    if(!!simulation.get()) {
+      simulation.remove();
+    }
+
+    await scope.run(sleep(5));
```
wKich, line 35: I think it could safely removed

.changes/config.json
```diff
@@ -48,6 +48,15 @@
...
       "publish": false,
       "releaseTag": false
+    },    
+    "@simulacrum/auth0-cypress": {
+      "path": "./integrations/cypress",
+      "manager": "javascript",
+      "dependencies": [
+        "@simulacrum/auth0-simulator"
+      ],
+      "publish": false,
```
cowboyd, line 58: Are we wanting to publish this?

packages/server/src/schema/resolvers.ts
cowboyd: Agree, let's remove it.

packages/server/src/schema/resolvers.ts
```diff
@@ -20,12 +21,19 @@ export interface CreateSimulationParameters {
...
     let { atom, scope, newid } = ctx;
 
-    let id = newid();
+    let id = options.key ?? newid();
+
     let simulation = atom.slice("simulations", id);
 
+    if(!!simulation.get()) {
+      simulation.remove();
+    }
```
cowboyd, line 33: Do we still need this functionality now that simulations are being fully destroyed after each test? I can still see having this as a feature, but maybe as a separate PR. 

In both cases though, we need to do the whole dance here that we do in `destroySimulation()` to make sure that the old simulation is torn down fully before we create the new one.

If we do move forward with this feature, I was thinking we should call it `uniqueKey` or `singletonKey`

PR #115, opened 2021-08-05
Transparently pass through person scenario

cowboyd:
## Motivation
Originally, so that the auth0 simulator could run as a standalone, it included the person scenario by itself. To do this, it needed to augment the person scenario by allocating also an email so that the person could be identified inside Auth0. Later on,  when the capability to pass parameters to any scenario was added, the wrapper was not expanded to pass the `params` down to the underlying scenario.

## Approach

This just passes the person scenario through un-altered with no wrapper.


dagda1 (approved):
👍

PR #116, opened 2021-08-05
use `email` in the id token data, not `mail`

cowboyd:
## Motivation

Id token data has the `email` field, not `mail` field: https://auth0.com/docs/api/authentication#get-user-info

This breaks an app if it is using the token data for the actual email.

## Approach

fix typo.

### DIFF
--- a/.changes/email-typo.md
+++ b/.changes/email-typo.md
@@ -1,0 +1,4 @@
+---
+"@simulacrum/auth0-simulator": patch
+---
+fix malformed token that had `mail` field, not `email` field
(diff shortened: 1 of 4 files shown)


cowboyd:
@dagda1 what do you think is the best way to test this?

dagda1:
> @dagda1 what do you think is the best way to test this?

I have added a test.  There already is a test to check for token fields.

dagda1 (approved):

jbolda (approved):

PR #117, opened 2021-08-09
examples: rules and cypress tests

jbolda:
## Motivation

This implements rules in our auth0 examples, and adds cypress tests. The cypress tests serve as a helpful example for a possible integration one might use, but also provides some guards as we work in the monorepo.

## Approach

Added cypress with interactor based tests, and running it in CircleCI.


jbolda:
There seems to be an issue with repeated hitting of the auth0 simulator. I skipped the tests for the moment so we aren't overloading this PR. I think it makes sense to get this merged in, and we can address other issues using these as confirmation.

examples/nextjs-with-auth0-react/cypress/integration/auth.spec.js
```diff
@@ -0,0 +1,124 @@
...
+  });
+
+  describe('login from the homepage', () => {
+    beforeEach(() => cy.visit('/'));
+
+    it('login is visible', () => {
+      cy.expect(Button('Login').exists());
+    });
+
+    it.skip('logs in user', () => {
```
dagda1, line 57: IMO I don't think we can merge this until all tests pass and nothing is skipped.

cowboyd: I agree

jbolda: I was trying to keep the scope of this PR tight and avoiding rabbit holes because I thought it would be helpful to get this in and useful for other PRs.

Alright, we can address this then. I don't know what is causing this, but it may be resolved in the other PRs. It seems like similar behavior? I will see if I can free up a bit of time to dig into this here or in the other PR.

jbolda:
Well, I don't know how I ran it before and got it to pass, but removing the skips again (which I think I might have lost when I rebased), the tests are failing. I think it is related to the global issues of cypress that we are using atoms to deal with in the plugin.

Sounds like we might want to adjust how these examples are configured since adding the plugins and recent changes have apparently superseded the tests here and made the example configuration confusing. Going to mark this as draft until it we decide how we want to proceed.

jbolda:
Seems that our needs have diverged from this PR. Going to just go ahead and close it.

PR #121, opened 2021-08-13
upgrade past @effection/core dist issue

jbolda:
## Motivation

In restructuring our package building in effection, `@effection/core` had a version missing `dist` assets. This bumps all effection related deps past this to prevent the issue from occurring downstream.

## Approach

Bumped all to `@latest` and pinned everything as well.


cowboyd (approved):
Looks good. We'll want to move this back to a pin once we get to 1.0 on effection

PR #122, opened 2021-08-13
covector publish when package doesn't exist

jbolda:
## Motivation

The existing command for the covector checks if a package has been published and the version. If the package has yet to be published, this command throws an error which breaks the publish sequence. This changes adjust the command to not throw an error and attempt to publish.

## Approach

A little extra bash.

### Alternate Designs

We want to make this more robust, and fetch via the API. This way we could check specifically for a 404 error. This likely makes sense after an effection upgrade and implementing presets. I can work on that next week, but have been focusing on simulacrum this week and haven't gotten to it. Don't want to hold up our publishing.

### DIFF
--- a/.changes/config.json
+++ b/.changes/config.json
@@ -3,7 +3,7 @@
   "pkgManagers": {
     "javascript": {
       "version": true,
-      "getPublishedVersion": "npm view ${ pkgFile.pkg.name } version",
+      "getPublishedVersion": "npm view ${ pkgFile.pkg.name } version || echo \"not published\"",
       "publish": "npm publish --access public"
     }
   },


cowboyd:
Is it not better to fail straight up than to not publish and succeed?

jbolda:
> Is it not better to fail straight up than to not publish and succeed?

This is only for the check to see if we need to publish a specific package. The current functionality is if the output of that command matches the version number, then we skip publishing for that package. Otherwise we attempt to publish.  The issue we are running into is that the command throws if it doesn't exist. This change just prevents it from throwing.

As I'm typing this out though, I realized we could literally write a tiny script and do `node checkIfPublished.js` with effection v2 and respond specifically to a 404 like you suggested.

jbolda:
@minkimcello what do you think of using a script like this? this might be a nice in-between from a one-line bash to a full on preset.

minkimcello:
> @minkimcello what do you think of using a script like this? this might be a nice in-between from a one-line bash to a full on preset.

I assume this is a temporary fix until similar functionality is built into covector?

Is this a workaround for when we want to publish the initial package to npm for the first time?

.changes/config.json
```diff
@@ -3,7 +3,7 @@
   "pkgManagers": {
     "javascript": {
       "version": true,
-      "getPublishedVersion": "npm view ${ pkgFile.pkg.name } version",
+      "getPublishedVersion": "node ../../.scripts/checkIfPublished.mjs ${ pkgFile.pkg.name } ${ pkg.tag ? pkg.tag : 'latest' }",
```
minkimcello, line 6: Shouldn't this be `node ../.scripts`?

jbolda:
> I assume this is a temporary fix until similar functionality is built into covector?
> 
> Is this a workaround for when we want to publish the initial package to npm for the first time?

Yes to both, but I imagine that it will be used verbatim in covector when we add the presets.

.changes/config.json
jbolda: Covector runs the commands with the `cwd` in the `./packages/*` directory for each package. So we need to go up two folder levels to the root.

minkimcello: Does this mean this won't work if any of the packages are two folders in?

jbolda: I think you mean it will _only_ work for packages that are two folders in. We happen to follow that in this repository though so it should work for the moment.

.scripts/checkIfPublished.mjs
```diff
@@ -0,0 +1,22 @@
...
+function* packageExists(name, tag) {
+  let request = yield fetch(`https://registry.npmjs.com/${name}`);
+  let response = yield request.json();
+  if (request.status === 404) {
+    return "not published";
+  } else if (request.status < 400) {
+    return response["dist-tags"][tag];
+  } else {
+    throw new Error("request error");
+  }
```
cowboyd, line 21: The `fetch` operation completes once all headers have been received, so no need to pull the body and parse as JSON unless we're going to need it. Also, because the body might be invalid json on a `404` this might be more correct.

```suggestion
  let request = yield fetch(`https://registry.npmjs.com/${name}`);
  if (request.status === 404) {
    return "not published";
  } else if (request.status < 400) {
    let response = yield request.json();
    return response["dist-tags"][tag];
  } else {
    throw new Error("request error");
  }
```

jbolda: ooh, yea that makes a lot of sense. Thanks!

jbolda:
@cowboyd are we OK with this as a first step forward until we can give more attention to it in covector?

cowboyd (approved):

PR #123, opened 2021-08-16
add bin script to enable npx auth0-simulator via #113

jbolda:
## Motivation

Adding a change file so the `npx auth0-simulator` shows up in the changelog.


taras (approved):
👍

PR #130, opened 2021-10-05
wait until simulation server is fully stopped

### LINKED ISSUE #127: auth0 simulation effection scope issue
The simulation server and auth0 integration have methods to create a simulation and destroy it. These methods appear to work in isolation, but when run repeatedly, it appears that the `destroy` tears down the `create` in the next step.

It can be reproduced on both the examples PR and on the cypress plugin (so seemingly not cypress specific). If you remove the `.skip` on the [example PR](https://github.com/thefrontside/simulacrum/pull/117/files#diff-60f18fa0040e05c68f7767b2da10cbd27756c050d2cdf5a3038bd9c2fd285384R71). On the [cypress plugin PR](https://github.com/thefrontside/simulacrum/pull/114/files#diff-ea9aa8c0538c8bf7648caa80df89a300e35402cbff7e2cf6ed300915783aad73R54-R112), running the tests should reproduce the issue.

wKich:
## Motivation

Fix #127 

## Approach

On `destroySimulation` request server starts closing asynchronously, at that moment it's still being able to accept incoming requests, so even if the new server is created, the new requests go to the old one. We don't wait until the server is fully stopped. So before sending `destroySimulation` response back, we wait for the simulation to be halted and the server is closed.


packages/server/src/http.ts
```diff
@@ -61,7 +61,15 @@ export function createServer(app: Application, options: ServerOptions): Resource
         try {
           yield;
         } finally {
-          server.close();
+          let { future, produce } = createFuture<void>();
```
cowboyd, line 64: Let's use `resolve()` and `reject()` in future instead since they're simpler. Also, we may deprecate `produce()` at some point.

packages/server/src/schema/resolvers.ts
```diff
@@ -26,6 +26,10 @@ export const createSimulation: Resolver<CreateSimulationParameters, SimulationSt
     let id = newid();
     let simulation = atom.slice("simulations", id);
 
+    if(!!simulation.get()) {
+      simulation.remove();
+    }
+
```
cowboyd, line 32: Why is this necessary?

cowboyd (commented):
Except for a few minor issues, looks very good!

packages/server/src/schema/resolvers.ts
wKich: It's from cypress-example branch, I can remove it.

packages/server/src/http.ts
wKich: In v0 the old version of effection, there is no `resolve/reject`

wKich:
@cowboyd done

PR #135, opened 2021-10-07
add basic logging middleware

dagda1:
## Motivation

We need more visibility as to what is going on in simulators like auth0 which are annoying to debug.

## Approach

There is a basic middleware component in `@simulacrum/server` that simulators can use to log HTTP requests.

It can take 1 or more logging function.  I've add one logging function to auth0 that writes to the affection inspector:

```ts
app: createHttpApp()
      .use(createLoggingMiddleware(function * (message) {
        yield label({ name: 'auth0-logging-middleware', log: message });
      }))
```

It should be easy to add other loggers like `console.log` for quicker debugging.

### DIFF
--- a/packages/server/src/middleware/console-logger.ts
+++ b/packages/server/src/middleware/console-logger.ts
@@ -24,13 +24,18 @@ function safeJSONStringify(o: Record<string, unknown>): string {
 
 export const consoleLogger: Middleware = function *(req, res) {
   console.log(`-----------------------------------------------
-[${getFormattedDate()}] 
+[${getFormattedDate()}]
+
 ${req.method.toUpperCase()}
+
 ${req.originalUrl ?? req.url}
+
 query
 ${safeJSONStringify(req.query ?? {})}
+
 body
 ${safeJSONStringify(req.body ?? {})}
+
 ${res.statusCode}
 -----------------------------------------------`);
 };


packages/server/src/middleware/http-logging-middleware.ts
```diff
@@ -0,0 +1,27 @@
...
+
+    let logMessage = log({ date, method, req, status });
+
+    for(let log of loggers) {
+      if(isGeneratorFunction(log)) {
+        yield log(logMessage);
+        continue;
+      }
+
+      log(logMessage);
```
cowboyd, line 24: Let's just make all loggers operations. We ran afoul of that before when trying to do that with streams so that you could use them with normal fat-arrow functions, and in the ended it turned out being better to just have everything be normalized. A generator will work even if there are no yields in it.

Also, I think we should pass in the `request and response` to the logger so that if the string isn't enough it can do its own thing.

cowboyd (commented):
Looks good. As mentioned in the comment, I'd just as soon remove the complexity of checking if something is a generator function because 
1.  An operation can something other than a generator function
2. The logger type`(log: String) => void` does not overlap with generator function, and we should strive to keep our typing coherent

I think the type of Logger should be:

```ts
interface Logger {
  (log: String, request: Request, repsonse: Response): Operation<void>;
}
```

PR #137, opened 2021-10-12
Upgrade to Effection 2.0

cowboyd:
## Motivation

Effection 2.0 is finally here. A stable version means we get to use `^` dependencies and experience a lot less churn!

## Approach

Bump them `package.json`


dagda1 (approved):
🚀  !!!

PR #154, opened 2021-10-19
Ldap processor

dagda1:
## Motivation

Add a basic ldap processor.

## Approach

Implementation is very basic and it would need to be tested more thoroughly.  Works for the basic backstage example.

### DIFF
--- a/packages/ldap/src/index.ts
+++ b/packages/ldap/src/index.ts
@@ -1,6 +1,6 @@
 import { createServer, InvalidCredentialsError, NoSuchObjectError, OperationsError, Server } from 'ldapjs';
 import { LDAPOptions } from './types';
-import type { SimulationState, Simulator } from '@simulacrum/server';
+import { SimulationState, Simulator } from '@simulacrum/server';
 import { ResourceServiceCreator } from '@simulacrum/server';
 import dedent from 'dedent';
 import { person } from '@simulacrum/server';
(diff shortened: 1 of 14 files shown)


packages/ldap/src/index.ts
```diff
@@ -0,0 +1,174 @@
...
+  port: 389
+};
+
+interface UserData{
+  id: string | number;
+  email: string;
+  password: string;
+}
+
+export function createLdapService<T extends UserData>(ldapOptions: LDAPOptions, state: Slice<SimulationState>): ResourceServiceCreator {
```
cowboyd, line 22: We should use consistent acronym naming. E.g. `createLDAPService`

packages/ldap/src/index.ts
```diff
@@ -0,0 +1,174 @@
...
+        };
+
+        server = createServer({ log });
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        server.search(baseDN, function (req: any, res: any, next: any) {
+          console.log(dedent`--- User Search ---
+          dn:     ${req.dn.toString()}
+          scope:  ${req.scope}
+          filter: ${req.filter.toString()}
+          `);
```
cowboyd, line 57: This should be behind a debug flag?

dagda1: we definitely could do that.

There is a real lack of good logging in simulacrum as a whole, I think we should have logging on by default while things are still taking shape.

This information is useful and you would use it as config for an application.

It can be really hard to find out why something is not working and I don't find much in the state, inspector or anywhere really so I end up logging stuff.

My shout would be to have more logging on by default and not less.

PR #159, opened 2021-11-01
Initial set up of docusaurus

jorgelainfiesta:
## Motivation

Set up a place where we can start writing out Simulacrum's documentatino

## Approach

I'm proposing "Graph-based [test data management(https://cloud.google.com/architecture/devops/devops-tech-test-data-management)" as tagline for Simulacrum. It's up for discussion, but it seems that the people who we've talked to see the value in that space. Maybe more than test data is like test services and their data, but I wanted to use a term that is familiar in shift-left

Brought over Docusaurus as we use it in other projects. I tried to upgrade from `2.0.0-beta.3` to `2.0.0-beta.8` but the maintainers added a feature, with a note that reads 'this is not idea' about the TOC generation that indeed is not a good idea as it relies on a single class in the nav (accessing a global context). Our classes are generated by vanilla-extract so the feature doesn't work. We can upgrade in a later version, nothing too valuable to us has been added in these minor versions.


cowboyd (approved):
I say we build it incrementally.

PR #160, opened 2021-11-25
Add debug option to createSimulation to add express middleware logging

dagda1:
## Motivation

Add debug option to the cypress createSimulation command to add express middleware logging.

## Approach

e.g.

```js
it("should log in and log out", () => {
  cy
    .createSimulation({ ...auth0Config, debug: true })
```

## Screenshots

out


cowboyd:
How does this interact with the server debug option?

dagda1:
> How does this interact with the server debug option?

It sets the `$debug` value in `createSimulation`.

I really don't like that errors are hidden and there is no way a user can know there are any unless they look at the state.  

That means they have to know the errors are in the state.

That means they will have to have read some documentation, and I don't believe we have documented this.

Errors without an error message or stack trace are one guarantee that users won't use your product.

We should raise errors and not rely on the user having read some doc that I don't think exists right now.

cowboyd (approved):

PR #161, opened 2021-12-23
One cypress command per file and nest examples under nextjs

dagda1:
## Motivation

There were a few things that I wanted to tidy before doing any more work on the cypress plugin.

The first was to make sure all the auth0 `scope` fields had the same value:

`AUTH0_SCOPE="openid profile email offline_access"`.

Next, I have always found it difficult to get my head around which example was which as the 2 directories had very similar names:

- nextjs-with-auth0-react
- nextjs-with-nextjs-auth0

This PR nests the 2 examples under a `examples/nextjs`  parent directory

```
|_ examples
   |_ nextjs
      |_ auth0-react
      |_ nextjs-auth0
```
Lastly, I've also split all the cypress commands into separate files for each command before doing any more work on this as it was becoming too big for 1 file.

commands


integrations/cypress/cypress/support/commands/logout.ts
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    client.destroySimulation(simulation).then(() => {
+      Cypress.log({
+        name: 'simulacrum-logout',
+        displayName: 'simulacrum-logout',
+        message: 'simulation destroyed'
+      });
+
+      atom.slice(Cypress.spec.name).remove();
```
taras, line 38: Is this synchronous? or do we need to wait for this operation to finish?

dagda1: that is synchronous

taras: Got it.

taras (approved):
👍

cowboyd (approved):

PR #162, opened 2022-01-04
Get user tokens

dagda1:
## Motivation

This PR enables the auth0 cypress plugin to run against both `nextjs-auth0` and `auth0-react`.

## Approach

Chuck brought [this package](https://github.com/sir-dunxalot/cypress-nextjs-auth0) to my attention which currently requires a real auth0 account.

The repo uses a slightly different approach that does not require `getTokenSilently` and also provides a helper to encrypt the `nextjs-auth0` session in a cypress cookie that was the root of the problems I had before bringing our version in line with this code.

It also makes use of the [openid connect endpoint /userinfo](https://connect2id.com/products/server/docs/api/userinfo) which we probably could get away with not using but I've implemented it for completeness.

Chuck said he knows the author and it might be an idea to reach out to him after using his approach.

I'm also now running the tests against both the `/nextjs-auth0` and `auth0-react` example repos.

### TODOs and Open Questions

I still have a bug when running cypress in the browser and if the tests are kicked off again after a successful run. 

I will fix that on a separate PR.


cowboyd (approved):

taras (approved):
👍

PR #164, opened 2022-01-06
Rerun cypress tests

dagda1:
## Motivation

Prior to this PR, running the cypress plugin in the browser was pretty unusable and would definitely have made this not right for prime time.

Every time a user pressed save or refreshed the browser, the simulation crashed into an unrecoverable state.

## Approach

The fix is to ensure that there is a no existing simulation before creating a new one:

```ts
export const makeCreateSimulation = ({ atom, getClientFromSpec }: MakeCreateSimulationOptions) => (options: CreateSimulation) => {
  return cy.logout().then(() => {
```
## Screenshots

https://user-images.githubusercontent.com/118328/148383468-436eeed7-1d6c-4ff4-a22f-dd315a29f90c.mp4

### DIFF
--- a/.changes/rerun-cypress-tests.md
+++ b/.changes/rerun-cypress-tests.md
@@ -1,0 +1,4 @@
+---
+"@simulacrum/auth0-cypress": minor
+---
+Destroy simulation before creating a new one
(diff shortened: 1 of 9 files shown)


cowboyd (approved):

taras (approved):
Looks solid

dagda1:
> Looks solid

For the very first time, it actually feels like something I would like other people to try 

> Looks solid

For the first time, it actually does.

PR #165, opened 2022-01-10
create defaults for auth0-cypress environment variables

dagda1:
## Motivation

When testing the @simulacrum/auth0-cypress package against a `create-react-app` package, it requires the `auth0ClientSecret` and `auth0CookieSecret` variables to be set in the `cypress.env.json` file.

They are not relevant when dealing with a SPA that renders in the client and is not calling an api that subsequently calls auth0.

## Approach

Centralise how the config is loaded and supply default values for the server side specific environment variables.

### DIFF
--- a/integrations/cypress/cypress/support/commands/login.ts
+++ b/integrations/cypress/cypress/support/commands/login.ts
@@ -50,8 +50,6 @@ export const makeLogin = ({ atom }: MakeLoginOptions) => () => {
             cy.task<string>('encrypt', payload).then((encryptedSession) => {
               log('successfully encrypted session');
 
-              console.dir({ encryptedSession });
-
               cy.setCookie(sessionCookieName, encryptedSession);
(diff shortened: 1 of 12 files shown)


integrations/cypress/cypress/support/commands/login.ts
```diff
@@ -49,6 +50,8 @@ export const makeLogin = ({ atom }: MakeLoginOptions) => () => {
             cy.task<string>('encrypt', payload).then((encryptedSession) => {
               log('successfully encrypted session');
 
+              console.dir({ encryptedSession });
+
```
cowboyd, line 54:
```suggestion
```

cowboyd (commented):
Was the `console.dir` intentional?

dagda1:
> Was the `console.dir` intentional?

definitely not.  cheers

PR #172, opened 2022-01-13
Auth0 cypress provider

dagda1:
## Motivation

Fixing the cypress plugin to work with `nextjs-auth0` broke the previous `auth0-react` code which I am guessing would be the main use case.

## Approach

Both `nextjs-auth0` and `auth0-react` have different authorization flows that was descending into a sea of horrible conditionals.

I have created separate `login` and `logout` cypress commands for each flow.

```ts
const providerCommands: Commands = {
  nextjs_auth0: {
    login: makeNextLogin,
    logout: makeNextLogout
  },
  auth0_react: {
    login: makeReactLogin,
    logout: makeReactLogout
  },
} as const;
```

Another problem was that the `nextjs-auth0` authorization flow uses the `auth0.js` and `auth0-react` uses `@auth0/auth0-react`.

I decided against trying to shoe horn both solutions into 1 sdk and I've used dynamic imports to avoid both packages getting bundled.


taras (approved):
👍

PR #174, opened 2022-01-19
Simplify createSimulation and destroySimulation

dagda1:
## Motivation

Creating and destroying simulations as side effects of state changes was proving problematic and the code was slightly opaque.

## Approach

Simplify things significantly by adding `createSimulation` and `destroySimulation` functions to the graphql context and call them from the resolvers in a more imperative manner.


packages/server/src/operation-context.ts
```diff
@@ -0,0 +1,70 @@
...
+          yield;
+        } finally {
+          simulationsMap.delete(simulationId);
+          slice.remove();
+        }
+      });
+
+      simulationsMap.set(simulationId, simulationTask);
+
+      await scope.run(slice.filter(({ status }) => status === 'running').expect());
```
cowboyd, line 53: can we put all of this in a single operation? that way, we only call one operation. I try to enter Effection world at as few points as possible, that way it makes error handling easier down the road because there is only one boundary:

```suggestion
      await scope.run(function*() {
        let simulationTask = yield scope.spawn(function* () {
          yield createSimulation(slice, simulators);

          try {
            yield;
          } finally {
            simulationsMap.delete(simulationId);
            slice.remove();
          }
        });        
        simulationsMap.set(simulationId, simulationTask);
        yield slice.filter(({ status }) => status === 'running').expect()
      });
```

cowboyd (approved):

PR #176, opened 2022-01-22
Auth0 cypress readme

dagda1:
## Motivation

The code snippet in the `@simulacrum/auth0-cypress` package is wrong and if copied will cause an error when the plugin loads.

## Approach

Fix the import statements in the REAMDE to ensure they point to the `encrypt` export that will not load all the commands before the `Cypress` global loads:

```ts
import { encrypt } from '@simulacrum/auth0-cypress/encrypt';

export default (on) => {
  on('task', { encrypt });
};
```


taras (approved):
👍

PR #177, opened 2022-01-26
Added new tags and description for simulators

taras:
## Motivation

I was linking to NPM packages for a Reddit post and found that our packages don't have good descriptions and don't have tags that would make them show up.

## Approach

I'm adding tags like this,

```json
    "mock",
    "mocking",
    "stubbing",
    "integration testing"
```

Also improving description for simulators and linking to simulators from README.md. Add a link to the blog post in Auth0 simulator.

BTW, I don't like these tags but we want people to be able to find them.


dagda1 (approved):
👍

PR #181, opened 2022-02-20
Eslint consitent types

dagda1:
## Motivation

Apply `@typescript/consistent-type-imports` across the code base.

## Approach

use `eslnt --fix`


taras (approved):
👍

PR #183, opened 2022-03-14
Rules tests

dagda1:
## Motivation

To fix a bug that occurs after a custom rule has added claims to the access token.

The claims were being added to the idToken and not the access token.

## Approach

Apply the result of the modification to the access token that is returned to the client.


taras (approved):
🚢

PR #184, opened 2022-03-15
Allow LDAP server to be standalone a resource

cowboyd:
## Motivation
In the article about running ingestion tests for backstage, we need to run a simulated LDAP server outside the context of simulacrum. The primary reason for this is that we do not need yet to introduce the concept of simulacrum as its own thing. It just complicates the world.

## Approach 
This just splits the LDAP simulator into two layers, the server resource, created by `createLDAPServer` and the service resource that adapts it into the Simulacrum context `createLDAPService()`. The user data is represented as an `Iterable` which is the lowest level JS format for specifying a list of values. It could be an arry, or it could be something more.

### Examples

```ts
let server = yield createLDAPServer({
  port: 3890,
  users: [{ email: 'admin@user.org', password: 'admin' }]
});
```

### DIFF
--- a/.changes/standalone-ldap-resource.md
+++ b/.changes/standalone-ldap-resource.md
@@ -1,0 +1,5 @@
+---
+"@simulacrum/ldap-simulator": minor
+---
+add low-level `createLDAPServer()` resource that lets you embed an LDAP server
+into any Effection program as an resource.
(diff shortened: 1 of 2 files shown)


dagda1 (approved):
👍

PR #186, opened 2022-03-15
Optionally output logs from LDAP simulator

cowboyd:
## Motivation
The LDAP simulator outputs a lot of useful debug output, however there is no way to silence this output and so when demoing test output, it can be difficult to demonstrate what is signal and what is noise.

## Approach
This adds a `log` flag to the server options, and won't print anything unless that flag is present

### DIFF
--- a/.changes/ldap-logging.md
+++ b/.changes/ldap-logging.md
@@ -1,0 +1,4 @@
+---
+"@simulacrum/ldap-simulator": patch
+---
+add `log` option to LDAP simulator options to enabled/disable logging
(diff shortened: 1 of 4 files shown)


dagda1 (approved):
Is logging off by default?  I think I would have it on by default for new users.   If things are not working, it is good to have some feedback of any sort.

I was going to say the `log` option should be added to the README but I never added one 🤦

taras (approved):
👍

cowboyd:
> Is logging off by default? I think I would have it on by default for new users. If things are not working, it is good to have some feedback of any sort.

@dagda1 I turned logging back on by default and it is disabled in the tests.

PR #187, opened 2022-03-15
make iat field epoch time

dagda1:
## Motivation

The iat field in both the idToken and the accessToken issued from `/oauth/token` is not currently in unix timestamp format.

## Approach

Convert iat field into a unix timestamp.


taras (commented):
It looks good, but diff is bigger than I would expect. Does it need a package-lock.json updated?

dagda1:
> It looks good, but diff is bigger than I would expect. Does it need a package-lock.json updated?

I've reverted it

taras (approved):
👍

taras:
Merge at your leisure

PR #192, opened 2022-04-06
Use LDAP client as a resource in tests

cowboyd:
## Motivation
As part of [adding the root DSE handler](https://github.com/thefrontside/simulacrum/pull/193) to the ldap simulator, I found that my bindings were getting out of sync because the after each hook was not attached to the appropriate scope. Since this is Effection, and we can use resources that automatically teardown after every test, let's make this implicit, that way our tests can be more flexible.

## Approach
This takes out the raw usage of the LDAP client, and wraps it in a resource. Specifically, in order to get to the ldap client, you must first `bind()` and then you can execute commands such as `search`. `search()` itself has been implemented as a stream which makes working with it a lot nicer.

### DIFF
--- a/packages/ldap/test/helpers.ts
+++ b/packages/ldap/test/helpers.ts
@@ -41,6 +41,9 @@ export function createLDAPClient(url: string): LDAP {
         *init() {
           let client = createLDAPJSClient({ url });
 
+          // The mere attempt to `bind()` requires an `unbind()`,
+          // so we have to put our ensure block first because
+          // it must be called even in the event that `bind()` fails.
           yield ensure(() => new Promise<void>((resolve, reject) => {
             client.unbind(err => {
               if (err) {
--- a/packages/ldap/test/helpers.ts
+++ b/packages/ldap/test/helpers.ts
@@ -79,8 +79,7 @@ export function createLDAPClient(url: string): LDAP {
                 });
 
                 yield spawn(function*() {
-                  let error = yield once(response, 'error');
-                  throw error;
+                  throw yield once(response, 'error');
                 });
 
                 yield spawn(on<SearchEntryObject>(response, 'searchEntry').forEach(publish))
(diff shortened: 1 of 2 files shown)


packages/ldap/test/helpers.ts
```diff
@@ -1,18 +1,19 @@
...
 import type { Client } from '@simulacrum/client';
 import { createClient } from '@simulacrum/client';
 import type { Server, ServerOptions } from '@simulacrum/server';
 import { createSimulationServer } from '@simulacrum/server';
-
 export type { Client, Simulation } from '@simulacrum/client';
+import { createClient as createLDAPJSClient, SearchEntryObject, SearchOptions } from 'ldapjs';
 
-export function createTestServer(options: ServerOptions): Resource<Client> {
+export function createTestServer(options: ServerOptions): Operation<Client> {
```
cowboyd, line 10: Best practice is to always return an opaque `Operation`

packages/ldap/test/helpers.ts
```diff
@@ -1,18 +1,19 @@
...
   return {
     *init(scope: Task) {
       let server: Server = yield createSimulationServer(options);
       let { port } = server.address;
       let client = createClient(`http://localhost:${port}`);
-      scope.run(function*() {
+      yield scope.spawn(function*() {
```
cowboyd, line 16: We should always use `scope.spawn()` since it is an operation rather than a synchronous method.

dagda1: What about an eslint for things like this?

dagda1: eslint could keep you developers like me on the straight and narrow in effection

cowboyd: @dagda1 That would be amazing actually. Another option is to straight up deprecate `Task#run()` in favor of having a some sort of `runWithin(operation, task)` In case you need to spawn a task synchronously from outside of an operation.

PR #193, opened 2022-04-06
Implement Root DSE search for LDAP Simulator

cowboyd:
> Depends on #192 

## Motivation
Every LDAP Server must have a [Root DSE][1] to provide metadata about itself. This allows clients to distinguish between particular vendors. Among others, Backstage uses the root dse to determine if it is talking to Microsoft Active Directory and customize its behavior accordingly.

## Approach
This adds the root DSE search which is just an "empty" entry with the zero leng string as its DN.

[1]: https://ldap.com/dit-and-the-ldap-root-dse/

### DIFF
--- a/.changes/ldap-root-dse.md
+++ b/.changes/ldap-root-dse.md
@@ -1,4 +1,4 @@
 ---
-@simulacrum/ldap-simulator: minor
+"@simulacrum/ldap-simulator": minor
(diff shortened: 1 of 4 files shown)


dagda1 (approved):
I've added #194 to improve the typing.  So I'm going to approve

PR #196, opened 2022-04-08
Simplify the LDAP schema

cowboyd:
## Motivation
There are some difficulties integrating the LDAP simulator in with the Backstage processors and providers. The `UserData` that feeds the LDAP server is required to have an `id` attribute and an `email` attribute, both of which are not part of any common ldap schema. Furthermore, it does not send common user attributes as part of a search result. So, for example, Backstage (as well as most common LDAP clients), when connecting to an OpenLDAP server, will be looking for both the `entryDN` virtual attribute and the `entryUUID` attribute. Backstage, for example, uses `entryUUID` as the value of the all-important `metadata.name` Entity field.

## Approach

This aligns the `user data` with the minimum number of attributes. Since `cn` is used to authenticate, we use `cn` in the user data api (also, the LDAP convention for the email field is `mail` anyway, and not `email`). This also makes send the the `dn` back to the client more straightforward since the `dn` is comprised of the `cn`,`baseDN`.

Note that this is a breaking change, and will require user data to be mapped differently for currently running these simulators.

There is actually quite a bit of variance in the attribute naming conventions that LDAP servers use, and so there really isn't a "one-size fits all". For example, there really is no guarantee that `entryUUID` is an attribute. I think we can mitigate this by specifying a "vendor" flag to show 

## Learning

There are some good resources for learning LDAP... not something that I'm good at yet, but I was able to come up to basic, basic speed with the following resources:

* https://ldap.com
* https://ldapwiki.com
and to a lesser extent, 
* https://openldap.org

In the future, it would be nice if we could run an actual open LDAP server embedded in-process using WASM.

### DIFF
--- a/.changes/align-ldap-schema.md
+++ b/.changes/align-ldap-schema.md
@@ -1,0 +1,4 @@
+---
+"@simulacrum/ldap-simulator": minor
+---
+feed ldap server uuid, cn directly to the low-level server api
--- a/packages/ldap/src/index.ts
+++ b/packages/ldap/src/index.ts
@@ -15,8 +15,8 @@ const DefaultOptions: Partial<LDAPOptions> = {
 };
 
 interface UserData {
-  id: string | number;
-  email: string;
+  uuid: string;
+  cn: string;
   password: string;
 }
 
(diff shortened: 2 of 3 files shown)


taras (changes requested):
Since this is a breaking change, do we want to do a major bump?

cowboyd:
> Since this is a breaking change, do we want to do a major bump?

I think we're still in pre-release mode here and pre-1.0 at that, so I think it should be fine. We don't have any clients using this that I am aware of.

taras (approved):
👍 but @dagda1 should review it before it's merged.

dagda1 (approved):
looks good to me and I think this can actually help with typing.

I take it, I used `email` instead of `mail` originally? 

I have no idea why I did that.

cowboyd:
> I have no idea why I did that.

Probably because LDAP is hard to make any sense of in the first place and getting something working is the first step to having anything nice at all 😜🤣🤣. I still have barely a concept of how it works.

PR #197, opened 2022-04-08
Fix watch script

cowboyd:
## Motivation
The watch script was using 2.0 pre-release effection APIs, and so were breaking.

## Approach
This brings them into line. The good news is that they are much more concise.


taras (approved):
👍

PR #200, opened 2022-04-12
Prep API and README for testing article

cowboyd:
## Motivation
There is an article in the works that references the LDAP simulator as a solution to testing a backstage server. Unfortunately, we don't have a README, and so the [NPM package homepage](https://www.npmjs.com/package/@simulacrum/ldap-simulator/v/0.4.0) looks quite bad.

## Approach
This adds a basic readme and shows how to start an LDAP server from both a vanilla JavaScript context as well from an Effection context. There is a third way, which is via a simulacrum server, but this does not add that since it's a bit more fiddly, likely to change, and not relevant for the content of the blog post.

It also adds the actual vanilla javascript integration since it didn't exist before (it just wraps the resource in a run);

### DIFF
--- a/.changes/ldap-readme.md
+++ b/.changes/ldap-readme.md
@@ -1,0 +1,5 @@
+---
+"@simulacrum/ldap-simulator": minor
+---
+Add a README to the NPM package as well as an async api around an LDAP server
+resource
--- a/packages/ldap/README.md
+++ b/packages/ldap/README.md
@@ -12,6 +12,7 @@ environment, you can use promise-based API.
 
 ## Plain JavaScript
 
+```js
 import { runLDAPServer } from "@simulacrum/ldap-simulator";
 
 async function run() {
@@ -39,6 +40,7 @@ async function run() {
     await server.close();
   }
 }
+```
 
 ## Effection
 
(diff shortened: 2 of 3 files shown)


dagda1 (approved):
The README is really needed.

I'm going to approve but one thing that is missing, is how to actually get data into the ldap server.  An ldap server without data won't be much use.

If the blog post outlines this then you could just link to the blog post.

taras (approved):
👍

taras:
> I'm going to approve but one thing that is missing, is how to actually get data into the ldap server. An ldap server without data won't be much use.

[It does actually show](https://github.com/thefrontside/simulacrum/pull/200/files#diff-725bc5f16f5c388a97e083059f595e91962bd4e57e2ad3eabd9d928b4cbf0085R24) passing data via users property when starting the server, but it's not called out in the description so it's easy to miss. 

We might want to add an API for this in the future, something like this.

```ts
const ldap = yield runLDAPServer();
yield ldap.createUser({ ... });
```

dagda1:
> [It does actually show]

I mean there is nothing in the README

taras:
I think it is, but the code formatting is broken

image

I just pushed a commit to fix it.

PR #215, opened 2022-09-29
Decouple Auth0 server from simulacrum

cowboyd:
## Motivation
Sometimes you want the freedom to just embed an Auth0 server directly into Node without having to pay the overhead of managing a simulacrum process. This pattern is already in play with the LDAP server, and it actually greatly simplifies the simulacrum server code itself. Instead of defining endpoints and then having simulacrum create the server, we just say "hey, start a server, and just let me know what the port and protocol are".

## Approach
This enables us to use the same simulator in both the simulacrum environment and also in any Node environment of our choosing by applying the same strategy to the Auth0 service as to the LDAP Service. The low level server accepts its list of people as an iterable, and any other information it may need from the environment. For example, rather than using the `Slice` of the simulation directly, it uses an adapter class similar to the `Map` API. That way it is easy to satify and adapt to.

Once this is complete, we can delete the LegacyResourceCreator interface and any code that requires it.

### DIFF
--- a/.changes/auth0-standalone.md
+++ b/.changes/auth0-standalone.md
@@ -1,0 +1,5 @@
+---
+"@simulacrum/auth0-simulator": minor
+---
+now exports a `createAuth0Server` operation which can be used directly without
+starting a Simulacrum server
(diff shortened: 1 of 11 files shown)


dagda1 (approved):
looks good to me.

cowboyd:
@dagda1 There's something wrong with the tests. Do you think you could help me debug it today?

dagda1:
> @dagda1 There's something wrong with the tests. Do you think you could help me debug it today?

Of course.  Just let me know when

PR #218, opened 2022-10-27
Auth0 `/userinfo` `access_token`

jbolda:
## Motivation

When implementing the auth0 simulator in Backstage, we ran into a situation where the authorization header was not set, but the `access_token` was passed as a query parameter. This PR takes it into consideration.

## Approach

Check for the authorization header, then fall back to the query parameter, `access_token`, and throw an assertion error if neither are present on the `/userinfo` endpoint.


dagda1 (approved):
🚢

PR #221, opened 2022-10-31
pass user data at /oauth/token

jbolda:
## Motivation

The `/oauth/token` endpoint appears to pass the user data through in the Auth0 examples. Mirror this functionality.

## Approach

I inspected a test response in the rules runner within Auth0, and this was the data that was input.


packages/auth0/src/handlers/auth0-handlers.ts
```diff
@@ -201,7 +201,31 @@ export const createAuth0Handlers = (store: Auth0Store, people: Iterable<Person>,
...
+          "identities": [
+            {
+              "provider": "auth0",
+              "user_id": "0123456789",
+              "connection": "Username-Password-Connection",
+              "isSocial": false
+            }
+          ]
+        }
+      */
```
cowboyd, line 220: I'm going to approve, but can you remove this before pushing?

cowboyd (approved):

packages/auth0/src/handlers/auth0-handlers.ts
jbolda: Sure, I can. I thought it might be nice to have an actual auth0 response, but perfectly fine removing it.

PR #222, opened 2022-11-01
Export standalone Auth0 creation function

cowboyd:
## Motivation
We want to be able to use the `createAuth0Server` directly so that we don't have to go through the ceremony of starting an entire Simulacrum server.

## Approach
This change exports the `Server` interface and the `createAuth0Server()` function.


dagda1 (approved):
🚢

jbolda (approved):

PR #227, opened 2022-11-16
allow setting the user.id in person simulator

jbolda:
## Motivation

There are cases where, in auth0, we are adding to the token context based on the user id. To help in simulating this situation, allow the `id` to be passed in as a parameter. This could result in conflicts, but I think it is on the them.

## Approach

The `id` as a parameter.


dagda1 (approved):
👍

PR #233, opened 2022-11-23
add express default error handler to auth0

dagda1:
## Motivation

Make the error handling logic in `/oauth/token` generic.

## Approach

Create an express default error handler.


packages/auth0/src/handlers/auth0-handlers.ts
```diff
@@ -162,28 +162,8 @@ export const createAuth0Handlers = (store: Auth0Store, people: Iterable<Person>,
...
-            .json({
-              error: {
-                name: error.name,
-                message: error.message,
-                stack: error.stack,
-              },
-            });
-        }
+      } catch (error) {
+        next(error);
```
jbolda, line 166: Do we need to have a try/catch here at all now? Will any thrown errors be automatically handled by the default error handler?

packages/auth0/src/error-handling-middleware.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    if (error.message.includes('::')) {
+      let errorMessage = error.message.slice(assertCondition.length);
+      errorCode = parseInt(errorMessage.slice(0, 3));
+      errorResponse = errorMessage.slice(5);
+    }
+
+    res.status(errorCode).send(errorResponse);
+  } else {
+    console.error(error);
```
jbolda, line 22: I accidentally forgot to remove this, but honestly.... I kind of don't mind it. It is nice to log out these unhandled conditions loudly rather than just silently failing with this stuff.

packages/auth0/src/handlers/auth0-handlers.ts
jbolda: (in voice chat) It seems that we need it in here. We might be able to deal with it more elegantly, but roll with it for now.

jbolda (approved):

PR #237, opened 2022-11-29
add stopgap debug in auth0-simulator

jbolda:
## Motivation

The middleware for debug logging is not working and requires a refactor. As a stopgap, adding a handful of `console.dir()` in the meantime.


taras (approved):
👍

PR #238, opened 2022-11-30
github API simulator

dagda1:
## Motivation

This is a bit rough, but the primary motivation is to decrease the surface area of the code we bring into a client's project.  

Importing a package is preferable to a tonne of code that a team has to maintain.

## Approach

The main export is a function that takes a `Graphgen` factory and starts a `@graphql/yoga` server.

```ts
export function startGithubApiSimulator(factory: GraphGen): void
```


taras (approved):
👍

PR #241, opened 2022-11-30
/login/callback should pass all wctx

jbolda:
## Motivation

The `/login/callback` is rather hard to inspect. It seems that we do need to pass the `client_id`, but I suspect there are other pieces we need as we extend the simulator. It seems pretty safe to pass the full object.


taras (approved):
👍

PR #242, opened 2022-11-30
auth0-simulator abstract logger on stopgap

jbolda:
## Motivation

> stopgaps never end up being stopgaps
> ~Paul

So giving it a bit of extra love before we :shipit:.


dagda1 (approved):
thank you!

packages/auth0/src/handlers/auth0-handlers.ts
```diff
@@ -29,6 +29,18 @@ export interface Auth0Store {
   set(nonce: string, session: AuthSession): void;
 }
 
+type LoggerArgs = Parameters<typeof console.dir>;
+
+const createLogger = (debug: boolean) => ({
+  log: (...args: LoggerArgs): void => {
+    if(debug) {
```
jbolda, line 36:
```suggestion
    if (!debug) {
```

PR #256, opened 2023-03-02
add the description mapping to Github API simulator

jbolda:
## Motivation

The description was not being mapped in the Github API simulator repository mapping.

## Approach

Added it to the mapping; having had tested the change downstream from the `node_modules`.


taras (commented):
👍

taras (approved):

PR #258, opened 2023-05-05
✨Add ability to extend the Github API Router with custom middleware

cowboyd:
## Motivation

Our implementation of the GitHub API is incomplete which is not suprising given that it is huge. However this poses a problem for someone using the API simulator. What if the API is missing something that they need? Their options are to submit a PR to this repo, or to just start the things they need on another port. This is doable, but it means that they have to manage two servers, and also that the simulation does not appear to be hosted at the same origin. That's an unecessary burden. We need a way to extend a simulator with new endpoints (and possibly new middlewares) without requiring an upstream change to this repository.

## Approach

This takes the simplest approach that could possibly work. It adds an "extension" property to the `ServerOptions` interface. It's just a function that, if present, will be invoked with the express router so that it can do whatever it wants to it.

### DIFF
--- a/.changes/extend-github-api-router.md
+++ b/.changes/extend-github-api-router.md
@@ -1,0 +1,4 @@
+---
+"@simulacrum/github-api-simulator": minor
+---
+Allow extenstion of github api simulator with new endpoints and middleware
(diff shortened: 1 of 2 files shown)


jbolda (approved):
🎉

PR #263, opened 2023-05-26
fix github api repository query name undefined

jbolda:
## Motivation

A `repository` query would fail due to a destructured `name`. This fixes the reference and adds an additional check for matching `nameWithOwner`.

## Approach

The first arg is the request context. The variables required are within the second arg passed to the repository query. Make this adjustment, and also use the `owner` field for improved matching.


cowboyd (approved):
👍🏻

cowboyd (approved):
👍🏻

PR #265, opened 2023-05-26
Github API simulator fields isArchived and defaultBranchRef

jbolda:
## Motivation

We have `isArchived` and `defaultBranchRef` missing.

## Approach

Simply adding them.


taras (approved):
👍

PR #267, opened 2023-09-06
`repositoryOwner` GitHub API Simulator Resolver

jbolda:
## Motivation

The `repositoryOwner` query was missing. Adding it.

## Approach

We added the resolver and abstracted out the return into a switch statement to more easily return the values consistently across queries and respecting the nesting and pagination.


.circleci/config.yml
```diff
@@ -18,7 +18,7 @@ jobs:
           command: mkcert -install && mkcert localhost
           working_directory: ~/.simulacrum/certs
       - node/install:
-          node-version: 18.12.0
+          node-version: 18.17.1
```
cowboyd, line 21: Is this required?

package.json
```diff
@@ -32,8 +32,8 @@
     "npm": ">=9"
   },
   "volta": {
-    "node": "18.16.0",
-    "npm": "9.6.7"
+    "node": "18.17.1",
+    "npm": "10.0.0"
   },
```
cowboyd, line 37: Is this necessary for this change?

cowboyd (commented):
Looks good 👍

I'd like to keep changes orthogonal. Also, do we have test for our GitHub simulator?

.circleci/config.yml
jbolda: I bumped to the latest because CircleCI was [complaining about newer npm](https://app.circleci.com/pipelines/github/thefrontside/simulacrum/1279/workflows/44aa26eb-ee4d-44f5-bfb1-77e19b9e0fe0/jobs/1369) and would fail the run. Happy to switch to a more appropriate fix if you have suggestions.

package.json
jbolda: See https://github.com/thefrontside/simulacrum/pull/267#discussion_r1318876740

jbolda:
We currently don't have any tests for the Github API simulator. Personally, I think tests and a test suite should come in separate PRs. We also need a README.

cowboyd (approved):

PR #268, opened 2023-09-07
Don't report server started until listener callback is invoked

cowboyd:
## Motivation

The server resource was being reported as initialized before the server.listen() callback was invoked. This meant that you could have test code that relied on the server running, and awaited server initialization, but the server was not actually up and accepting connections.

## Approach
During server initialization, create a promise that is only ever resolved when the server.listen() callback is invoked.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,6 @@
 version: 2.1
 orbs:
-  node: circleci/node@4.3.0
+  node: circleci/node@5.1.0
 jobs:
   test:
     docker:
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -1,6 +1,6 @@
 version: 2.1
 orbs:
-  node: circleci/node@4.3.0
+  node: circleci/node@5.1.0
```
cowboyd, line 3: This was required to make the tests pass 🤷🏻‍♂️

jbolda (approved):

PR #276, opened 2024-05-25
covector fetch preset

jbolda:
## Motivation

Using the fetch preset in covector means we can remove our bespoke script that was doing the same.

## Approach

We adjusted the covector config and removed the script. This also allows us to remove the `npm ci` required for the script which will improve the speed to the `status` command.


frontsidejack:
No changes.

cowboyd (approved):

PR #285, opened 2024-08-28
foundation simulator delay response API

jbolda:
## Motivation

Add API to delay responses to improve "real API feel".

## Approach

Add middleware which takes an input or an environment variable to configure response delays.

### DIFF
--- a/packages/foundation/src/middleware/delay.ts
+++ b/packages/foundation/src/middleware/delay.ts
@@ -1,0 +1,61 @@
+import type { Request, Response, NextFunction } from "express";
+
+export function delayMiddleware(
+  delayResponses?: number | { minimum: number; maximum: number }
+) {
+  const delayMin =
+    typeof delayResponses === "number"
+      ? delayResponses
+      : delayResponses?.minimum;
+  const delayMax =
+    typeof delayResponses === "number" ? undefined : delayResponses?.maximum;
+
+  const minFromProcess = process.env.SIM_RESPONSE_DELAY_MIN
+    ? process.env.SIM_RESPONSE_DELAY_MIN
+    : process.env.SIM_RESPONSE_DELAY;
+  const min = minFromProcess ? parseInt(minFromProcess, 10) : delayMin;
+
+  const maxFromProcess = process.env.SIM_RESPONSE_DELAY_MAX;
+  const max = maxFromProcess ? parseInt(maxFromProcess, 10) : delayMax;
+
+  return async function delayHandler(
+    request: Request,
+    response: Response,
+    next: NextFunction
+  ) {
+    if (min || max) {
+      let timeoutDuration = calculateTimeoutDuration(min, max);
+      await new Promise<void>((resolve) => {
+        let timeoutHandle: NodeJS.Timeout | undefined = undefined;
+
+        const done = () => {
+          if (timeoutHandle) {
+            clearTimeout(timeoutHandle);
+          }
+          resolve();
+        };
+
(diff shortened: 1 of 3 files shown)


frontsidejack:
### Package Changes Through 209ffa55bfd3532b2771e3bfb03ef771d1a05eed
There are 1 changes which include @simulacrum/foundation-simulator with minor

<details>
<summary><i>Planned Package Versions</i></summary>

The following package releases are the planned based on the context of changes in this pull request.
| package | current | next |
  |----|----|----|
| @simulacrum/foundation-simulator | 0.1.0 | 0.2.0 |

</details>

[Add another change file through the GitHub UI by following this link.](https://github.com/thefrontside/simulacrum/new/foundation-delay?filename=.changes/change-pr-285.md&value=---%0A%22@simulacrum/foundation-simulator%22:%20patch%0A%22@simulacrum/auth0-simulator%22:%20patch%0A%22@simulacrum/ldap-simulator%22:%20patch%0A%22@simulacrum/github-api-simulator%22:%20patch%0A---%0A%0Afoundation%20simulator%20delay%20response%20API%0A)


---

packages/foundation/src/middleware/delay.ts
```diff
@@ -0,0 +1,61 @@
...
+  const delayMax =
+    typeof delayResponses === "number" ? undefined : delayResponses?.maximum;
+
+  const minFromProcess = process.env.SIM_RESPONSE_DELAY_MIN
+    ? process.env.SIM_RESPONSE_DELAY_MIN
+    : process.env.SIM_RESPONSE_DELAY;
+  const min = minFromProcess ? parseInt(minFromProcess, 10) : delayMin;
+
+  const maxFromProcess = process.env.SIM_RESPONSE_DELAY_MAX;
+  const max = maxFromProcess ? parseInt(maxFromProcess, 10) : delayMax;
```
cowboyd, line 19: These environment variables shouldn't be present inside a middleware, but instead should be passed in as configuration options. Environment variables can then be read and and inputs assigned at the point of entry (usually in main, or in a config module that is instantiated from main)

cowboyd (commented):
How do you configure the delay from within a test case? Maybe with a header? Or with a specific control API?

packages/foundation/src/middleware/delay.ts
jbolda: Aye, we do accept those inputs and I imagine that is going to be the main code path used. However, I think this specific instance might make sense to also support this. It is, in a way, a similar case to something like the `DEBUG=*` env var and `debug` crate.

For example, if I want to quickly test a "throttle" or slow connection type situation, I can use that env var and it works for _everything_. I am not hard set in supporting this directly, but it feels worthwhile.

jbolda:
> How do you configure the delay from within a test case? Maybe with a header? Or with a specific control API?

I haven't really ventured too far down controlling from tests. In part because I haven't had the need, but also I really just want my tests to run fast 🤣 .

It does feel worthwhile to support, but I am comfortable punting on it for now. I am not terribly keen on headers because then your production code paths might need to muddy themselves up with that header. Control API feels better, but I don't think we have enough in place to support that yet. What do you think?

packages/foundation/src/middleware/delay.ts
jbolda: We spoke and the overriding reason to remove the `process.env` is it will "win" and override any configuration passed in when (eventually) configured in a CLI. We want the CLI command args to take precedence, then `process.env`, then configuration.

cowboyd (approved):
Nice!

PR #312, opened 2025-06-15
starfx 0.14.x in foundation simulator

jbolda:
## Motivation

In upgrading to `starfx@0.14.x`, we can remove some of the dep tree that came in partially broken through peerDeps.

## Approach

Upgrade here, but more refined tweaks upstream.


jbolda:
Closing in favor of #324.

PR #313, opened 2025-07-10
Include email claim in access token when "email" specified as a scope

taylorreece:
## Motivation

When you request an access token from Auth0 and include both an `email` scope and an `aud`, Auth0 will include an `email` claim in both the access token _and_ ID token. Currently, the simulator only includes the email in the ID token.

This update checks if `email` is specified in the `scope`, and if so, includes an `email` claim in the access token.

## Screenshots

image

### DIFF
--- a/.changes/email-in-access-token.md
+++ b/.changes/email-in-access-token.md
@@ -1,0 +1,5 @@
+---
+"@simulacrum/auth0-simulator": patch
+---
+
+Update the auth0 simulator to include an email claim in the access token when email is present in scope.


jbolda:
Appreciate the PR!

I was going to mention that you can try it in the preview package, but it appears that I haven't updated the workflow to consider forks yet 😞 . Rather small change though so I am fine releasing it as a patch right away. Could you add a change file per the `./.changes/README.md`? (That is what [the normal comment](https://github.com/thefrontside/simulacrum/pull/312#issuecomment-2973527332) would say.)

jbolda (approved):

PR #315, opened 2025-08-14
[GH Sim] better handle nested graphql query data

jbolda:
## Motivation

We were getting errors from the `@octokit/graphql` client.

## Approach

- Add specific tests with both fetch and the `@octokit/graphql` client to ensure neither return errors.
- Fix existing errors.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -120,6 +120,7 @@
       "version": "7.28.4",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.27.1",
         "@babel/generator": "^7.28.3",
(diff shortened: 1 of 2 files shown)


packages/github-api/src/graphql/to-graphql.ts
```diff
@@ -1,108 +1,157 @@
...
+}
+
+export function deriveOwner(
+  simulationStore: ExtendedSimulationStore,
+  login: string
+) {
+  let [org] = simulationStore.schema.organizations
+    .selectTableAsList(simulationStore.store.getState())
+    .filter((o) => o.login === login);
+  console.dir(org, { depth: null });
```
cowboyd, line 37: We should remove this?

cowboyd (approved):

Issue #362, opened 2026-04-01
Replace `cosmiconfig` with `configliere` in auth0


jbolda:
We want to push CLI configuration handling forward in concert with the plan in #361.

- create new `configliere` based parser
- programmatic use expects a parsed config
- remove `cosmiconfig`
- export config helper for loading json file
- use config helper in standalone CLI to handle json

jbolda:
@a-kriya if you are still interested in the `http` in auth0, we could wrap it into this or follow up this work. Then we just add it to the config parser and handle all the downstream. `configliere` will handle all the env vars and config stuff.

a-kriya:
I was mainly interested in making running the tests simpler (without requiring specific system state). But you mentioned some of the tests must be run over https, if I understood correclty.

jbolda:
We can support configuration to allow it to run over `http`. I specifically wanted the tests to still be able to run over `https` in CI. Pass in something like `process.env.CI ? 'https' : 'https'`.

a-kriya:
Makes sense.

So, would you like me to tackle this ticket, or do I just work on making protocol/hostname configurable once you're done with this?

jbolda:
Either or both would be most appreciated! My main concern with your changes in the previous PR were that most of related to the work noted in this issue and not your original issue. Also, the ability to "toggle" the protocol started to leak into production code paths, and if we were going to do that, it felt best done intentionally through `configliere`.

a-kriya:
The README of latest configliere is outdated. It has references to `createParser` on a program, which doesn't exist in 0.3.0, I think. Having updated docs would help me put up a first draft of this migration.

PR #363, opened 2026-04-11
feat(auth0): switch from cosmiconfig to configliere (#362)

a-kriya:
## Motivation

Migrates from cosmiconfig to in-house configliere for parsing configuration options.

Adds a `protocol` config option to allow users to choose serving over `http` or `https`.

## Summary by CodeRabbit

* **New Features**
  * Auth0 simulator now supports CLI-driven configuration with `--help` and an explicit `start` command.
  * Configure the simulator using CLI flags, JSON files, environment variables, or programmatic options.
  * Startup output now displays the active server URL and default-user credentials.

* **Documentation**
  * Expanded the README with configuration guidance, supported options, and precedence rules.

* **Updates**
  * Updated the Auth0 client script used by the login page to version 9.32.0.


frontsidejack:
### Package Changes Through b4a2d6ffa3b0a44ef4ffd0e48a7f567353e67f47
No changes.

[Add a change file through the GitHub UI by following this link.](https://github.com/a-kriya/simulacrum/new/config?filename=.changes/change-pr-363.md&value=---%0A%22@simulacrum/server%22:%20patch%0A%22@simulacrum/foundation-simulator%22:%20patch%0A%22@simulacrum/auth0-simulator%22:%20patch%0A%22@simulacrum/github-api-simulator%22:%20patch%0A---%0A%0Afeat(auth0):%20switch%20from%20cosmiconfig%20to%20configliere%20(#362)%0A)


---

a-kriya:
@jbolda Will you be able to help test this, and make improvements. I can't login with the default user due to SSL errors even when setting `PROTOCOL=http`.

jbolda:
Hmm, how are you testing it out? If it uses the auth0 libs, those tend to throw errors if they aren't pointing at https. Is that the error you are getting? That is why we had the instructions to use `mkcert` on first run. I might expect the tests to pass at least without setting up that cert.

a-kriya:
I was running the start script, which executes `./example/index.mts`. Then going to the `/login` route and entering the default user credentials. Auth0 webAuth was hitting `https://localhost` because the domain passed to it did not specify protocol (`domain: new URL(serviceURL(req)).host`). I'll ignore that for now.

For the unit tests, I modified them in a separate commit to pass `protocol: 'http'` option, so they now all pass without depending on system state.

image

jbolda:
As a bit of a separate question, do you have some requirements that prevent you from installing a local dev-only cert? I was considering automating that process more, a few libs can help some with it, but it the ramifications of trying to do it automatically were unclear.

a-kriya:
Not really, `mkcert` is a dependency that hasn't been maintained for a few years. While I realize it might just be "complete" and doesn't require further work, I was just a bit wary of installing it and wanted to first figure out if the test suite can pass without it. And it did, so if it provides the same confidence it seems like an improvement because of a simpler setup.

Regardless, if you want to drop the third commit of this PR, please feel free.

packages/auth0/src/handlers/auth0-handlers.ts
```diff
@@ -43,7 +43,7 @@ export const createAuth0Handlers = (
   options: Auth0Configuration,
   debug: boolean,
 ): Record<Routes, RequestHandler> => {
-  let { audience, scope, clientID, rulesDirectory } = options;
+  let { audience, scope, clientId, rulesDirectory } = options;
```
jbolda, line 46: I think `clientID` was explicitly used as that aligns with Auth0 docs. Could we confirm that and maybe link it to their docs for future reference?

packages/auth0/src/views/login.ts
```diff
@@ -27,7 +27,7 @@ export const loginView = ({
           href="https://unpkg.com/tailwindcss@^2/dist/tailwind.min.css"
           rel="stylesheet"
         />
-        <script src="https://cdn.auth0.com/js/auth0/9.19.0/auth0.js"></script>
+        <script src="https://cdn.auth0.com/js/auth0/9.32.0/auth0.js"></script>
```
jbolda, line 30: Huh. I wonder if we make this configurable? We could set this value as the default, but it might be worth allowing folks to opt for a specific version?

packages/auth0/src/config/get-config.ts
```diff
@@ -1,56 +1,39 @@
...
+  options?: Partial<Auth0Configuration>,
+  args: string[] = [],
+): Auth0Configuration {
+  const envs = [{ name: "env", value: process.env as Record<string, string> }];
+  const values = [{ name: "options", value: options }];
+  const result = auth0Program.parse({ args, envs, values });
+
+  if (!result.ok) throw result.error;
+
+  return result.value.config;
```
jbolda, line 38: It looks like you have some good progress here. We will need to parse with a file as well. The `inject` is the new part of the API to help in handling it. [This example that uses the newer `inject` which we will need to load a config file](https://github.com/thefrontside/configliere/blob/main/examples/phased-commands.ts). After we have our parser, the return of that is effectively an object. That object is what we can pass directly to the `auth0` simulator.

packages/auth0/src/index.ts
```diff
@@ -21,19 +21,21 @@ export type Auth0Simulator = (args?: {
     extendRouter?: (router: Router, simulationStore: ExtendedSimulationStore) => void;
   };
   options?: Partial<Auth0Configuration>;
+  args?: string[];
 }) => FoundationSimulator<ExtendedSimulationStore>;
 
 export const simulation: Auth0Simulator = (args = {}) => {
-  const config = getConfig(args.options);
+  const config = getConfig(args.options, args.args);
```
jbolda, line 28: The slight adjustment here is that we don't do the config handling here. `args.options` should just be passed all of the data that we get from `config`. We won't parse in this function as that should be complete by the time we call this function. I don't think it will create a breaking change as before and after we only have an object. This distinction though means we can do more advanced parsing and pass in the result here.

packages/auth0/test/ci-smoke.test.ts
```diff
@@ -1,83 +1,51 @@
...
-          } catch (ignore) {
-            // ignore and retry
-          }
-          if (Date.now() - start > 30000) return Err(new Error("service did not start"));
-          yield* sleep(200);
-        }
-      },
-    },
-  });
-}
```
jbolda, line 46: Do we have a reason that we needed to remove the `useService()`?

packages/auth0/test/openid-handlers.test.ts
```diff
@@ -4,12 +4,12 @@ import type { FoundationSimulatorListening } from "@simulacrum/foundation-simula
...
 
 let basePort = 4401;
-let host = "https://localhost";
+let host = "http://localhost";
 let auth0Url = `${host}:${basePort}`;
 describe("openid routes", () => {
   let server: FoundationSimulatorListening<unknown>;
   beforeAll(async () => {
-    const app = simulation();
+    const app = simulation({ options: { protocol: "http" } });
```
jbolda: I still want to run this with `https` in CI. Can we make it a ternary or a flag? `vitest` might have a helper here too to allow us to run it in "https mode" or "http mode".

packages/foundation/src/store/index.ts
```diff
@@ -153,6 +153,10 @@ export function createSimulationStore<
     },
   );
 
+  let additionalTasks = [thunks.register];
+
+  let [schema, schemaInitialState] = generateSchemaWithInputSlices(inputSchema);
+
```
jbolda: I am curious if we moved this for a specific reason?

jbolda (changes requested):

packages/foundation/src/store/index.ts
a-kriya: I kept getting the following warning in the terminal when I was working on this last, but I can't reproduce anymore.

> [update] is being called before its thunk has been registered. Run `store.run(thunks.register)` where `thunks` is the name of your `createThunks` or `createApi` variable.

a-kriya: Will drop this commit.

packages/auth0/test/openid-handlers.test.ts
a-kriya: Alright, probably for another PR. I'll keep this change to confirm the config changes aren't affecting things, and to keep the focus on the migration in this PR.

packages/auth0/test/ci-smoke.test.ts
a-kriya: With the simplification, the test verified the same things, so this looked redundant to me. I realize I don't have the background.

packages/auth0/src/handlers/auth0-handlers.ts
a-kriya: The WebAuth instantiation in views/login.ts still specifies `clientID`, but using it as our config option was creating `--client-i-d` and `CLIENT_I_D` as the parameters.

packages/auth0/src/views/login.ts
a-kriya: Can do once the migration is settled. In most cases, users would be browsing the docs for the latest version and expect that API.

packages/auth0/src/index.ts
a-kriya: How would start.cjs look? Would `index.ts` export `getConfig` which `start.cjs` will call and pass to `simulation()`?

packages/foundation/src/store/index.ts
jbolda: I think I got a pin on this issue, and I fixed it in the other PR or upstream in starfx. Keep an eye on it though!

packages/auth0/test/openid-handlers.test.ts
jbolda: I was thinking we do something like this, then we can swap between them by running with a different CLI flag.
https://vitest.dev/guide/test-context.html#default-fixture-injected

Feel free to try that now or we can do it later. Definitely want `https` always in CI though.

packages/auth0/src/index.ts
jbolda: Yep! Or at least the config parser. `getConfig()` could probably be removed as you would want to handle commands and then call `simulation` only in some cases (instead of the early return that you have).

packages/auth0/src/handlers/auth0-handlers.ts
jbolda: Ahhh I see. @cowboyd do we have a good way to deal with this?

jbolda: I think we are going to have to put in an override API into `configliere` for this use case. I will do my best to whip that up here. I suspect that we are going to run into issues otherwise in getting it to work with auth0 libs if we don't follow their semantics.

a-kriya: Alright. Although I do think that this project should be able to use any naming format within itself, and only the interface layer with Auth0 needs to worry about using parameters that Auth0 accepts.

packages/auth0/src/config/get-config.ts
a-kriya: I need some help here. Would you mind taking this further in the direction that we want to go to?

packages/auth0/src/handlers/auth0-handlers.ts
jbolda: I don't inherently disagree. We want the incoming config to match whatever official auth0 accepts though. Also some of those variables are leaky into like the JWT if I recall. 

I did get the PR going though, so we can update that soon here.
https://github.com/thefrontside/configliere/pull/19

packages/auth0/src/config/get-config.ts
jbolda: Sure I can toss a commit here in a little bit.

packages/auth0/src/handlers/auth0-handlers.ts
a-kriya: Ah right, makes sense.

packages/auth0/src/config/get-config.ts
jbolda: Ok I pushed a commit. I think this is a good direction. We will still want to check the APIs to make sure we understand any breaking changes.

PR #373, opened 2026-07-08
Add Microsoft Entra ID (Azure AD) OIDC simulator

taras:
## Motivation

We want to run applications that authenticate against **Microsoft Entra ID**
(formerly Azure AD) locally — for development, previews, and integration tests —
without mock data and without changing the application's authentication source
code. Today we have the auth0 simulator but no Entra equivalent.

This adds `@simulacrum/entra-simulator` as a **drop-in replacement for the core
Entra OIDC authentication user flows**. An application only needs to change
*where* its identity provider lives (the authority/issuer URL); every core
sign-in flow then behaves as it would against real Entra.

## Approach

New package at `packages/entra`, modeled directly on `packages/auth0` and built
on `@simulacrum/foundation-simulator`.

- **Tenant-scoped v2.0 endpoints** mirroring real Entra: `/:tenant/oauth2/v2.0/{authorize,token,logout}`,
  `/:tenant/v2.0/.well-known/openid-configuration`, `/:tenant/discovery/v2.0/keys`,
  `/:tenant/discovery/instance` (covers MSAL's `/common/discovery/instance`), and a
  Graph-style `/oidc/userinfo`. The `:tenant` path segment is authoritative — it
  becomes the `tid` claim and issuer tenant so discovery `issuer` and issued
  tokens stay internally consistent.
- **Flows:** authorization-code + PKCE (S256/plain), refresh_token,
  client_credentials (app-only token with `roles`), and ROPC/password. Supports
  `response_mode` of `query`, `fragment`, `form_post`, and silent auth
  (`prompt=none`) via the session cookie.
- **Entra v2.0 token shapes:** id/access tokens carry `ver`, `iss`, `sub`, `aud`,
  `oid`, `tid`, `preferred_username`, `email`, `nonce`, `scp`/`roles`, `azp`, and
  the id_token header sets `kid` matching the JWKS so real client libraries
  verify signatures. Reuses the same dev RSA key as the auth0 simulator.
- Authorization codes and refresh tokens are encoded statelessly as base64url
  JSON (as auth0 does for refresh tokens), so the token endpoint needs no
  server-side session for the exchange. PKCE is verified per RFC 7636.
- README documents drop-in config for **MSAL**, **passport-azure-ad/NestJS**, and
  **next-auth**.

### Verification

- 19 tests pass (`vitest`, matching the auth0 sibling), covering discovery/JWKS/
  instance discovery, the full auth-code + PKCE flow, all grant types, userinfo,
  logout, and silent auth.
- Manually drove the built server end-to-end with `curl` as a real client would,
  including **cryptographically verifying the issued id_token against the
  published JWKS** with issuer + audience validation.
- `tsc`, `oxlint`, `oxfmt`, and the `tsdown` build (attw + publint) are all clean.

A changeset is included (`minor` bump for the new package).

## Summary by CodeRabbit

* **New Features**
  * Added a local Microsoft Entra ID (OIDC) simulator package with OpenID discovery, JWKS, AAD instance discovery, authorization, token, userinfo, and logout endpoints.
  * Supports authorization-code + PKCE, refresh tokens, client credentials, and password (ROPC) sign-in flows.
  * Added a new start/boot entrypoint plus an example server for quick local testing.

* **Documentation**
  * Added full package README and updated the main documentation to include the new simulator.

* **Tests**
  * Added integration tests covering discovery metadata, sign-in redirects and response modes, token issuance, userinfo, silent auth, refresh/client-credentials/password grants, and logout.


frontsidejack:
### Package Changes Through 56105c89b1b7288bb2f5c52ddca3e699d486ea13
There are 1 changes which include @simulacrum/entra-simulator with minor

<details>
<summary><i>Planned Package Versions</i></summary>

The following package releases are the planned based on the context of changes in this pull request.
| package | current | next |
  |----|----|----|
| @simulacrum/entra-simulator | 0.1.0 | 0.2.0 |

</details>

[Add another change file through the GitHub UI by following this link.](https://github.com/thefrontside/simulacrum/new/entra-simulator?filename=.changes/change-pr-373.md&value=---%0A%22@simulacrum/server%22:%20patch%0A%22@simulacrum/foundation-simulator%22:%20patch%0A%22@simulacrum/auth0-simulator%22:%20patch%0A%22@simulacrum/entra-simulator%22:%20patch%0A%22@simulacrum/github-api-simulator%22:%20patch%0A---%0A%0AAdd%20Microsoft%20Entra%20ID%20(Azure%20AD)%20OIDC%20simulator%0A)


---

taras:
Thanks @coderabbitai — triaged all 15 findings through a **developer-experience lens**, since this is a local dev tool that never runs in production (so security-only hardening is out of scope here).

**Addressed (commit 56105c8):**
- `utils.ts` — malformed authorization codes / refresh tokens now return `400 invalid_grant` instead of an opaque 500 + stack trace
- `entra-handlers.ts` — `autoPostForm` now HTML-escapes reflected values (framed as correctness: a legit `state`/`redirect_uri` with `"`/`&`/`<` was corrupting the posted `state`), and the redirect uses `&` when `redirect_uri` already has a query string
- `store/index.ts` — `extendStore` now forwards `logs`
- `package.json` — `types` points at the emitted `./dist/index.d.mts`
- `openid-handlers.ts` — dropped `jwt-bearer`/`implicit` from `grant_types_supported`
- `token.ts` — `auth_time` is preserved across refresh
- README — clarified `/oidc/userinfo` is mounted globally, not tenant-scoped
- Added tests for `fragment`/`form_post` modes, escaping, the separator fix, and malformed-grant 400s

**Intentionally not changed:**
- Session `httpOnly` and gating the error stack trace behind a debug flag — security-only concerns that don't apply to a localhost dev tool (the stack trace actively helps developers debug)
- README MD040 tags, tsconfig `.mts` include, `baseUrl` Host guard, and `const`-vs-`let` in the example (the repo's house style uses `let`) — low/no DX value

Note: the `logs`-forwarding and `types`-field issues also exist in the auth0 sibling; left as an optional follow-up to keep this PR scoped.

packages/entra/package.json
```diff
@@ -0,0 +1,90 @@
...
+    "@faker-js/faker": "^9.3.0",
+    "@simulacrum/foundation-simulator": "^0.8.0",
+    "assert-ts": "^0.3.4",
+    "base64-url": "^2.3.3",
+    "cookie-session": "^2.1.0",
+    "cors": "^2.8.6",
+    "cosmiconfig": "^9.0.0",
+    "express": "^5.2.1",
+    "html-entities": "^2.5.2",
+    "jose": "^5.9.6",
```
jbolda, line 77: Can we use the latest v6 here? It is on my list to update auth0 as well. (Maybe double check we can't update any of the others as well? Except `cosmiconfig` as we have some work already to use `confligliere` instead.)

packages/entra/package.json
```diff
@@ -0,0 +1,90 @@
+{
+  "name": "@simulacrum/entra-simulator",
```
jbolda, line 2: Noting so we don't forget, we discuss a possible name change.

packages/entra/package.json
```diff
@@ -0,0 +1,90 @@
...
+  "version": "0.1.0",
```
jbolda, line 3: Set this to v0 so the first bump is a minor to this version.

packages/entra/src/views/login.ts
```diff
@@ -0,0 +1,80 @@
...
+              value="${encode(query.login_hint ?? "")}"
+            />
+            <input
+              id="password"
+              name="password"
+              type="password"
+              autocomplete="current-password"
+              placeholder="Password"
+              required
+            />
```
jbolda, line 64: How are you using this and signing in? In my more recent custom client work, I ended up making this a `<select>` where you just choose a user from the store to sign in as. No need to remember username / password.

(We could probably do this in auth0 as well.)

jbolda (changes requested):

packages/entra/src/handlers/entra-handlers.ts
```diff
@@ -0,0 +1,251 @@
...
+      logger.log({ "/token": { body: { ...req.body, client_secret: "***" } } });
+      try {
+        let tokens = await createTokens({
+          simulationStore,
+          config,
+          issuer: issuerFor(req),
+          tenant: tenantParam(req),
+          body: req.body ?? {},
+        });
+        res.status(200).json(tokens);
```
jbolda, line 204: Does this endpoint support an `expires_in`? See https://github.com/thefrontside/simulacrum/blob/main/packages/auth0/src/handlers/auth0-handlers.ts#L208 in auth0.

jbolda (commented):
Are you doing anything with the session? Looks like you encode it in a cookie, then allow a refresh without checking session first? In auth0, we dump it in the store and use that to check if you have first signed in.
https://github.com/thefrontside/simulacrum/blob/main/packages/auth0/src/handlers/auth0-handlers.ts#L152-L158

